PR #907, opened 2016-08-31
Adds gearman background jobs for long running tasks

iturgeon:
NOTE - requires docker to be on the same issue branch

splits data/get_export to 2 endpoints, the api, and the new document download endpoint
updates file::render x-sendfile support to support multiple directories, not just media
adds Lib_Gearman for creating gearman clients and workers
updates notification to add support for data export notifications
baseurl now required in config so workers know how to build urls
new packages/materia/documents path for storing documents like data exports
data_export_request added to api
Perm::DOCUMENT constant added
New worker tasks added.  php oil r workers will set up a gearman worker process
adds gearman worker for build_export_data
adds rel="noopener noreferrer" to some links, need to continue doing more
Adds new Angular notification service to centeralize repeated code
Score export now uses api to request a file, gearman worker processes it, and a notification with a download link is created when its done

TODO: score export ui still needs some work, waiting for the global alert/modal
TODO: more noopenr links


iturgeon:
Moved out to my fork and issued a new draft PR #1237 1237

Issue #1188, opened 2019-01-02
Question ID Management needs an overhaul


clpetersonucf:
We're not strictly enforcing uniqueness of question IDs the way we probably should: for instance, importing questions doesn't scrub the existing question ID, and importing the same question multiple times is an easy way to cause collisions. We need to better strip/regenerate question and answer IDs in a number of places:
- JS: in engine/creatorcore, prior to being passed to the creator during question import
- PHP: when storing into question bank

We'll have to manage handling both existing question IDs and creation/handling of new question IDs.

Lastly, we should switch to using uuids in `find_questions` on the PHP side.

FrenjaminBanklin:
Refer to https://github.com/ucfopen/Materia-Server-Client-Assets/issues/37.

clpetersonucf:
Resolved via #1189

PR #1189, opened 2019-01-04
Replaces uniqids with uuids for question/answer ID generation

clpetersonucf:
Replaces the md5 hash generation with uuids when creating new IDs for questions and answers in qsets.

Refers to #1188 and and https://github.com/ucfopen/Materia-Server-Client-Assets/issues/37


iturgeon:
Should we target this to dev/5-ifirit

PR #1199, opened 2019-02-12
Allow for widgets to restrict who can publish instances.

FrenjaminBanklin:
Closes #1198.

Made a migration to track whether a widget is restricted for student creation or not.
Adjusted installer to account for a new install.yaml property corresponding to the new field.
Adjusted widget catalog lookup to omit author_only widgets when non-author users are looking at the catalog.
Fixed My Widgets page not properly disabling copy/delete buttons for users who lack appropriate access to a widget instance.

Essentially allows widgets to define themselves as 'non-student' widgets, meaning students cannot create new instances of them.

Release alongside https://github.com/ucfopen/Materia-Server-Client-Assets/pull/47.

PR #1203, opened 2019-02-22
Lock a widget instance in the creator rather than at the My Widgets screen.

### LINKED ISSUE #1202: Widget instance locking needs to occur when the creator is opened.
Currently a widget instance is locked as soon as somebody clicks the 'Edit' button in the My Widgets page.

Not only does this mean that if somebody can't actually edit a widget (i.e. #1198) it'll still lock the widget, but it also means that going straight to the create url for a widget bypasses the lock check entirely.

Potentially the creator could have a new API call 'start_edit_session' or something which locks the widget, gets publish privileges, etc. all in one go.

FrenjaminBanklin:
Closes #1202.

Relies on #1199.

Adjusts some communications between the front and back ends to determine who can edit widget instances and when.

### DIFF
--- a/fuel/app/classes/basetest.php
+++ b/fuel/app/classes/basetest.php
@@ -87,7 +87,7 @@ class Basetest extends TestCase
 		return \Materia\Widget_Manager::search($search)[0]->id;
 	}
 
-	protected function make_disposable_widget($name = 'TestWidget', $restricted = false)
+	protected function make_disposable_widget(string $name = 'TestWidget', bool $restrict_publish = false)
 	{
 		$user = $this->make_random_author();
 
@@ -103,7 +103,7 @@ class Basetest extends TestCase
 				'is_playable' => true,
 				'is_editable' => true,
 				'in_catalog' => true,
-				'restrict_publish' => $restricted,
+				'restrict_publish' => $restrict_publish,
 				'api_version' => 2,
 			],
 			'score' => [
--- a/fuel/app/classes/basetest.php
+++ b/fuel/app/classes/basetest.php
@@ -87,7 +87,7 @@ class Basetest extends TestCase
 		return \Materia\Widget_Manager::search($search)[0]->id;
 	}
 
-	protected function make_disposable_widget($name = 'TestWidget')
+	protected function make_disposable_widget($name = 'TestWidget', $restricted = false)
 	{
 		$user = $this->make_random_author();
 
@@ -103,6 +103,7 @@ class Basetest extends TestCase
 				'is_playable' => true,
 				'is_editable' => true,
 				'in_catalog' => true,
+				'restrict_publish' => $restricted,
 				'api_version' => 2,
 			],
 			'score' => [
@@ -134,6 +135,7 @@ class Basetest extends TestCase
 			'widget'          => $widget,
 			'is_student_made' => true,
 			'guest_access'    => true,
+			'published_by'    => $user->id,
 			'attempts'        => -1
 		]);
 		$demo_inst->db_store();
--- a/fuel/app/classes/basetest.php
+++ b/fuel/app/classes/basetest.php
@@ -87,7 +87,7 @@ class Basetest extends TestCase
 		return \Materia\Widget_Manager::search($search)[0]->id;
 	}
 
-	protected function make_disposable_widget(string $name = 'TestWidget', bool $restrict_publish = false)
+	protected function make_disposable_widget(string $name = 'TestWidget', bool $restrict_publish = false): \Materia\Widget
 	{
 		$user = $this->make_random_author();
 
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -139,7 +139,7 @@ class Api_V1
 			'guest_access'    => $is_student,
 			'attempts'        => -1
 		]);
-		if ( ! $is_draft && ! $widget->publishable_by(\Model_User::find_current_id()) ) return new Msg(Msg::ERROR, 'Widget can not be published by student!');
+		if ( ! $is_draft && ! $widget->publishable_by(\Model_User::find_current_id()) ) return new Msg(Msg::ERROR, 'Widget type can not be published by students.');
 
 		if ( ! empty($qset->data)) $inst->qset->data = $qset->data;
 		if ( ! empty($qset->version)) $inst->qset->version = $qset->version;
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -72,14 +72,12 @@ class Api_V1
 	 * @return object, contains properties indicating whether the current
 	 * user can edit the widget and a message object describing why, if not
 	 */
-	static public function widget_instance_edit_perms_verify($inst_id)
+	static public function widget_instance_edit_perms_verify(string $inst_id): \stdClass
 	{
 		$response = new \stdClass();
 		$response->msg = null;
 		$response->is_locked = true;
 		$response->can_publish = false;
-
-		$locked_by = \Cache::easy_get('instance-lock.'.$inst_id);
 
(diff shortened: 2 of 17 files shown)


fuel/app/classes/materia/widget/instance/manager.php
```diff
@@ -64,6 +64,22 @@ public static function get_all_for_user($user_id)
...
 	}
 
+	/**
+	 * Checks to see if the given widget instance is locked for the current user.
+	 *
+	 * @param inst_id widget instance id to check for lock
+	 *
+	 * @return bool whether the given instance is locked for the current user
+	 */
+	public static function locked_by_current_user($inst_id)
```
iturgeon:
```suggestion
	public static function locked_by_current_user(string $inst_id): bool
```

fuel/app/tests/api/v1.php
```diff
@@ -532,6 +545,88 @@ public function test_widget_instance_delete()
 		$this->assert_invalid_login_message($output);
 	}
 
+	public function test_widget_instance_edit_perms_verify()
```
iturgeon:
```suggestion
	public function test_widget_instance_edit_perms_verify(): void
```

fuel/app/tests/api/v1.php
```diff
@@ -532,6 +545,88 @@ public function test_widget_instance_delete()
...
+
+		// ======= AUTHOR ========
+		$this->_as_author();
+		$output = Api_V1::widget_instance_edit_perms_verify($instance->id);
+		$this->assertFalse($output->is_locked);
+		$this->assertTrue($output->can_publish);
+		$this->assertNull($output->msg);
+	}
+
+	public function test_widget_publish_perms_verify()
```
iturgeon:
```suggestion
	public function test_widget_publish_perms_verify(): void
```

fuel/app/tests/widgets/widget.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * @group App
+ * @group Widget
+ * @group Materia
+ */
+
+class Test_Widget extends \Basetest
+{
+	public function test_publishable_by()
```
iturgeon:
```suggestion
	public function test_publishable_by(): void
```

fuel/app/classes/materia/api/v1.php
```diff
@@ -68,6 +68,48 @@ static public function widget_instance_delete($inst_id)
...
+	 * user can edit the widget and a message object describing why, if not
+	 */
+	static public function widget_instance_edit_perms_verify($inst_id)
+	{
+		$response = new \stdClass();
+		$response->msg = null;
+		$response->is_locked = true;
+		$response->can_publish = false;
+
+		$locked_by = \Cache::easy_get('instance-lock.'.$inst_id);
```
iturgeon: $locked_by isn't used here?

fuel/app/classes/materia/api/v1.php
```diff
@@ -68,6 +68,48 @@ static public function widget_instance_delete($inst_id)
...
+	/**
+	 * @return bool, true if the current user can publish the given widget instance, false otherwise.
+	 */
+	static public function widget_publish_perms_verify($widget_id)
+	{
+		if (\Service_User::verify_session() !== true) return Msg::no_login();
+		if ( ! Util_Validator::is_pos_int($widget_id)) return Msg::invalid_input($widget_id);
+
+		$widget = new Widget();
+		if ( $widget->get($widget_id) == false) return Msg::invalid_input('Invalid widget type');
```
iturgeon: This error message seems wrong - this error would pop up if the widget instance wasn't found right?

fuel/app/classes/materia/api/v1.php
```diff
@@ -125,6 +167,7 @@ static public function widget_instance_new($widget_id=null, $name=null, $qset=nu
 			'guest_access'    => $is_student,
 			'attempts'        => -1
 		]);
+		if ( ! $is_draft && ! $widget->publishable_by(\Model_User::find_current_id()) ) return new Msg(Msg::ERROR, 'Widget type can not be published by students.');
```
iturgeon: I'd do this as soon as it's possible - move up to after `widget->get()` is called

fuel/app/classes/basetest.php
```diff
@@ -103,6 +103,7 @@ protected function make_disposable_widget($name = 'TestWidget')
 				'is_playable' => true,
 				'is_editable' => true,
 				'in_catalog' => true,
+				'restrict_publish' => $restricted,
```
iturgeon: suggest renaming `$restricted` to `$restrict_publish`

fuel/app/classes/materia/api/v1.php
FrenjaminBanklin: Nah, this one specifically is checking to see if a widget (non-instance) exists - hence 'widget type'. Not really sure about 'widget type' as the vocabulary, but 'type' is probably still a better word than 'template'.

fuel/app/classes/materia/api/v1.php
```diff
@@ -68,6 +68,46 @@ static public function widget_instance_delete($inst_id)
...
+
+		if ( ! Util_Validator::is_valid_hash($inst_id)) $response->msg = Msg::invalid_input($inst_id);
+		else if (\Service_User::verify_session() !== true) $response->msg = Msg::no_login();
+		else if ( ! static::has_perms_to_inst($inst_id, [Perm::FULL])) $response->msg = Msg::no_perm();
+		else if ( ! ($inst = Widget_Instance_Manager::get($inst_id))) $response->msg = new Msg(Msg::ERROR, 'Widget instance does not exist.');
+
+		//msg property only set if something went wrong
+		if ( ! $response->msg)
+		{
+			$response->is_locked = ! Widget_Instance_Manager::locked_by_current_user($inst_id);
```
clpetersonucf: The use case here seems to suggest we're checking if the instance is locked _for_ the current user, not _by_ the current user.

fuel/app/classes/materia/widget/instance/manager.php
```diff
@@ -64,6 +64,22 @@ public static function get_all_for_user($user_id)
...
 	}
 
+	/**
+	 * Checks to see if the given widget instance is locked for the current user.
+	 *
+	 * @param inst_id widget instance id to check for lock
+	 *
+	 * @return bool whether the given instance is locked for the current user
+	 */
+	public static function locked_by_current_user(string $inst_id): bool
```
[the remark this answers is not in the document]
FrenjaminBanklin: It can technically be used either way, since 'true' means the widget is locked BY the current user and 'false' means the widget is locked FOR the current user, but you're right that the doc code should match the vocabulary.

fuel/app/classes/materia/api/v1.php
```diff
@@ -68,6 +68,46 @@ static public function widget_instance_delete($inst_id)
 		return $inst->db_remove();
 	}
 
+	/**
+	 * @return object, contains properties indicating whether the current
+	 * user can edit the widget and a message object describing why, if not
+	 */
+	static public function widget_instance_edit_perms_verify(string $inst_id): \stdClass
+	{
+		$response = new \stdClass();
```
iturgeon: Style wise - I'd move this initialization below the validations that follow

fuel/app/classes/materia/api/v1.php
iturgeon: Looks ok to me now

fuel/app/classes/materia/widget/instance/manager.php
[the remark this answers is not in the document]
iturgeon: name seems ok to me now.  current docs mean what it does to me.

fuel/app/classes/materia/api/v1.php
FrenjaminBanklin: This method will always return a `stdClass` and the validations all potentially modify `$response->msg`, so it'd have to be defined beforehand. The alternative is to have a `$msg` that we define at the top and have all the validations set that instead, then define `$response` afterwards, but that seems like it'd be unnecessary.

PR #1221, opened 2019-05-31
polish docker scripts and clean up release builds

iturgeon:


iturgeon:
Still needs to split out the deploy script on travis so it only runs when theres a tag (release)

Issue #1217, opened 2019-05-16
For non-scoring widgets, send a 100% participation scores


iturgeon:
In order to support use of non-scoring widgets for participation scoring - we could just send back scores when the widget is launched.

When a widget is lti launched and provides a score passback url  check to see if the widget is a non-scoring widget.  If it is, immediately send back a 100.  If not, let the widget's logic do what it does.

Add a dialog when using the picker to inform the user that they have chosen a non-scoring widget, and what it will do.

clpetersonucf:
Implemented via #1226, individual widgets must implement the new engine call.

PR #1226, opened 2019-06-26
issue/1217 - submit participation scores

weining-li:
Added unscorable log type to be used along with: https://github.com/ucfopen/Materia-Server-Client-Assets/pull/63

#1217

### DIFF
--- a/fuel/app/migrations/047_add_score_unscorable_to_log.php
+++ b/fuel/app/migrations/047_add_score_unscorable_to_log.php
@@ -1,0 +1,24 @@
+<?php
+namespace Fuel\Migrations;
+class Add_score_unscorable_to_log
+{
+	public function up()
+	{
+		\DBUtil::modify_fields(
+			'log',
+			[
+				'type' => ['type' => 'enum', 'constraint' => "'BUTTON_PRESS', 'ERROR_GENERAL', 'ERROR_TIME_VALIDATION', 'KEY_PRESS', 'SCORE_ACTIVITY_FROM_CLIENT', 'SCORE_FINAL_FROM_CLIENT', 'SCORE_QUESTION_ANSWERED', 'SCORE_WIDGET_INTERACTION', 'SCORE_UNSCORABLE', 'WIDGET_CORE_INIT', 'WIDGET_END', 'WIDGET_LOAD_DONE', 'WIDGET_LOAD_START', 'WIDGET_LOGIN', 'WIDGET_PLAY_REQ', 'WIDGET_PLAY_START', 'WIDGET_RESTART', 'WIDGET_START', 'WIDGET_STATE', 'DATA'", 'null' => true]
+			]
+    );
+  }
+
+	public function down()
+	{
+		\DBUtil::modify_fields(
+			'log',
+			[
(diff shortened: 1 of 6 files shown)


weining-li:
Added a migration for `SCORE_UNSCORABLE`.

There is a test in the Materia Server Client Assets side of the code. Does another test need to be added here? I couldn't find an example of the test for other types like `SCORE_QUESTION_ANSWERED`

FrenjaminBanklin:
The test you wrote in the MSCA is only making sure the front end half of the process works. There are some [tests for the API](https://github.com/ucfopen/Materia/blob/master/fuel/app/tests/api/v1.php) on the server side which you should probably look into updating to deal with the new log type.

Admittedly the state those tests are in leaves a lot to be desired, but typically any time we add something new to the back end we also add or adjust testing to make sure the new feature works as expected. There's a block that has a few tests for adding play logs already which starts on [line 628](https://github.com/ucfopen/Materia/blob/dc95058f860aa0e9472bf7309906e12745ee69ed/fuel/app/tests/api/v1.php#L628), but ideally a test to make sure this new log type works as expected would have its own block.

iturgeon:
@clpetersonucf Make sure to note this change has a migration.  It happens to be on the gnarliest table in the system, which may incur some downtime.

fuel/app/migrations/047_add_score_unscorable_to_log.php
```diff
@@ -0,0 +1,24 @@
...
+		\DBUtil::modify_fields(
+			'log',
+			[
+				'type' => ['type' => 'enum', 'constraint' => "'BUTTON_PRESS', 'ERROR_GENERAL', 'ERROR_TIME_VALIDATION', 'KEY_PRESS', 'SCORE_ACTIVITY_FROM_CLIENT', 'SCORE_FINAL_FROM_CLIENT', 'SCORE_QUESTION_ANSWERED', 'SCORE_WIDGET_INTERACTION', 'SCORE_UNSCORABLE', 'WIDGET_CORE_INIT', 'WIDGET_END', 'WIDGET_LOAD_DONE', 'WIDGET_LOAD_START', 'WIDGET_LOGIN', 'WIDGET_PLAY_REQ', 'WIDGET_PLAY_START', 'WIDGET_RESTART', 'WIDGET_START', 'WIDGET_STATE', 'DATA'", 'null' => true]
+			]
+    );
+  }
+
+	public function down()
+	{
```
FrenjaminBanklin: This throws an SQL error, `Uncaught exception PDOException: 01000 - SQLSTATE[01000]: Warning: 1265 Data truncated for column 'type' at row...`

If there are any rows in the `log` table where the value is `SCORE_UNSCORABLE`, the modify query will error out since it's changing which values are allowed to exist in that column while there are rows in the table with a value that would no longer be allowed.

In order to avoid this, you'd have to make sure you remove any rows in the table with the invalid value OR modify them to have a valid value - either way, you'd have to do that prior to the modify query.

For example, to delete the would-be-erroneous rows:
```suggestion
\DB::delete('log')->where('type', 'SCORE_UNSCORABLE')->execute();
```

weining-li:
This is turning out to be a little more complicated than I initially thought. :O

Issue #1158, opened 2018-09-17
Default collaboration level for new collaborators is 'nothing'.


FrenjaminBanklin:
![screen shot 2018-09-17 at 10 22 45 am](https://user-images.githubusercontent.com/1275983/45629167-bff78b00-ba63-11e8-954a-342a0944a94e.png)

This is technically valid, but it seems nonsensical. Should probably default to 'View Scores'.

clpetersonucf:
Resolved via #1231

PR #1231, opened 2019-07-01
issue/1158 - default collaboration level

weining-li:
Using ng-options for the dropdown list to get rid of the blank option.

#1158

### DIFF
--- a/fuel/app/themes/default/partials/my_widgets.php
+++ b/fuel/app/themes/default/partials/my_widgets.php
@@ -82,7 +82,7 @@
 									<span class="owner">Full</span>
 									<span class="undo">Removed <a href="#">Undo</a></span>
 
-									<select ng-disabled="selected.shareable==false" tabindex="0" id="perm" class="perm" ng-model="$scope.accessLevels[0]" ng-options="accessLevel as accessLevel.text for accessLevel in accessLevels" ng-change="checkForWarning(collaborator)">
+									<select ng-disabled="selected.shareable==false" tabindex="0" id="perm" class="perm" ng-model="accessLevels[0]" ng-options="accessLevel as accessLevel.text for accessLevel in accessLevels" ng-change="checkForWarning(collaborator)">
 										</select>
 
 									<a ng-if="selected.shareable" tabindex="0" class="remove-expiration" role="button" ng-click="removeExpires(collaborator)" ng-show="collaborator.expires">X</a>
--- a/fuel/app/themes/default/partials/my_widgets.php
+++ b/fuel/app/themes/default/partials/my_widgets.php
@@ -82,11 +82,12 @@
 									<span class="owner">Full</span>
 									<span class="undo">Removed <a href="#">Undo</a></span>
 
-									<select ng-disabled="selected.shareable==false" tabindex="0" id="perm" class="perm" ng-model="collaborator.access" ng-change="checkForWarning(collaborator)" ng-options="accessLevel.value as accessLevel.text for accessLevel in accessLevels">
-										</select>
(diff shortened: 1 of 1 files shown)


fuel/app/themes/default/partials/my_widgets.php
```diff
@@ -82,10 +82,8 @@
 									<span class="owner">Full</span>
 									<span class="undo">Removed <a href="#">Undo</a></span>
 
-									<select ng-disabled="selected.shareable==false" tabindex="0" id="perm" class="perm" ng-model="collaborator.access" ng-change="checkForWarning(collaborator)">
-										<option value={{ACCESS.FULL}} ng-selected="collaborator.access == ACCESS.FULL" >Full</option>
-										<option value={{ACCESS.VISIBLE}} ng-selected="collaborator.access == ACCESS.VISIBLE" >View Scores</option>
-									</select>
+									<select ng-disabled="selected.shareable==false" tabindex="0" id="perm" class="perm" ng-model="collaborator.access" ng-change="checkForWarning(collaborator)" ng-options="accessLevel.value as accessLevel.text for accessLevel in accessLevels">
```
FrenjaminBanklin: The `ng-options` directive seems like maybe it's too restrictive for our purposes here. Would recommend using an an `ng-repeat` and defining the options separately.

```suggestion
<select ng-disabled="selected.shareable==false"
	tabindex="0"
	id="perm"
	class="perm"
	ng-model="collaborator.access"
	ng-change="checkForWarning(collaborator)">
	<option ng-repeat="level in accessLevels"
		value="{{level.value}}">
		{{level.text}}
	</option>
</select>
```

This probably won't solve the original problem, which was that newly added collaborators' access level was defaulted to an empty third option. I believe that's a problem that will need to be solved in the client-side code.

fuel/app/themes/default/partials/my_widgets.php
```diff
@@ -83,12 +83,11 @@
...
 
 									<select ng-disabled="selected.shareable==false" tabindex="0" id="perm" class="perm" ng-model="collaborator.access" ng-change="checkForWarning(collaborator)">
-										<option value={{ACCESS.FULL}} ng-selected="collaborator.access == ACCESS.FULL" >Full</option>
-										<option value={{ACCESS.VISIBLE}} ng-selected="collaborator.access == ACCESS.VISIBLE" >View Scores</option>
+										<option ng-repeat="accessLevel in accessLevels" value={{accessLevel.value}} ng-selected="collaborator.access == accessLevel.value">{{accessLevel.text}}</option>
 									</select>
 
 									<a ng-if="selected.shareable" tabindex="0" class="remove-expiration" role="button" ng-click="removeExpires(collaborator)" ng-show="collaborator.expires">X</a>
-									<span class="expires">Expires: </span><input ng-disabled="!selected.shareable" type="text" class="exp-date user{{::collaborator.id}}" ng-model="collaborator.expiresText" readonly="true" />
+									<span class="expires">Expires: </span><input ng-disabled="!selected.shareable" type="text" class="exp-date user{{::collaborator.id}}" ng-model="collaborator.expiresText" readonly="true" /><p>{{n}}</p>
```
clpetersonucf: What's the purpose behind the `<p>{{n}}</p>`? My guess is it's a debug element, but want to be 100% sure.

FrenjaminBanklin: Seems likely, since there's no `$scope.n` in the Angular half of this.

Just get rid of the debug stuff and this should be ready to go.

weining-li: Oh yeah, it was just for debugging. I must have overlooked it. Sorry.

clpetersonucf:
Looks good now, just needs @FrenjaminBanklin's approval.

clpetersonucf:
:+1:

PR #1236, opened 2019-07-09
adds post event messages so apps containing materia can listen in

iturgeon:
This has bearing on what Obojobo can do when listening to Materia.  Keep in mind this is already being used for scores in Obojobo Classic, but Obojobo Next could use greater insight by listening to the messages.

Worth noting these post messages aren't super secure, mostly useful for state change information.  Score information or user authenticity should be cryptographically signed to fully trust it.


clpetersonucf:
Stale, closing, Materia appears to be compatible with OboNext without this feature.

PR #1238, opened 2019-07-09
Allow widgets to save their current play state so a play can be resumed later.

FrenjaminBanklin:
Adds support to Materia platform to receive and save play states from widgets, and recall those states when a user tries to play the same widget prior to completing it.


clpetersonucf:

Issue #1216, opened 2019-05-08
Inline help on my widgets to inform users of integration options


iturgeon:
There are instructors that are still exporting scores and importing them into Canvas.  This led me to the thought that we should try to make sure these people are aware of their options (they may have been using materia before it was able to sync)

Right now we can see 2 places where this would be useful as just in time help - Near the link on the mywidgets page, and on the score exporting dialog.

iturgeon:
> Gonna just make sure the documentation is up to date and link out to that?
via @FrenjaminBanklin 

I'm thinking that as a minimum.  Though I could value in at least naming the options inline in case new options become available. Iif it's just a link, new options wouldn't get any attention.

clpetersonucf:
Davis implemented half of this, but I still think it's important to include a better explainer on the "Share with your Students" section as well - likely bundled with a more comprehensive redo of the My Widgets page.

PR #1242, opened 2019-07-15
#1216 Inline help on my widgets to inform users of integration options

Goff-Davis:
- Added some info for users regarding exporting scores to canvas. Gave a link to documentation.
- Will only show when user has selected the option to export all scores.

For issue #1216


Goff-Davis:
See PR # 77 in Materia Server Client Assets for style changes

Issue #1256, opened 2019-09-27
Swap "Player Guide"/"Creator Guide" tab positions to match widget page


clpetersonucf:
The guide links are listed like this in on the page for each widget:
<img width="387" alt="Screen Shot 2019-09-27 at 10 55 39 AM" src="https://user-images.githubusercontent.com/1268547/65779222-65cce380-e115-11e9-9b32-88b869c0be95.png">

But the tabs are listed like this on the guide pages themselves:
<img width="306" alt="Screen Shot 2019-09-27 at 10 55 27 AM" src="https://user-images.githubusercontent.com/1268547/65779243-6f564b80-e115-11e9-8a8d-28b3b4c1f92f.png">

And it's driving me a little crazy every time I notice

clpetersonucf:
Completed via #1257

PR #1257, opened 2019-10-01
#1256 Swap positions of player and creator guide links

Goff-Davis:
- Swapped player and creator guide links on the widget details page to match the tabs


Goff-Davis:
See #1256

clpetersonucf:
Moved merge target to dev/7

PR #1261, opened 2019-10-09
issue/76 delay export options dialogue until $scope.selected.scores.list exists

weining-li:
https://github.com/ucfopen/Materia-Server-Client-Assets/issues/76

### DIFF
--- a/fuel/app/themes/default/partials/my_widgets.php
+++ b/fuel/app/themes/default/partials/my_widgets.php
@@ -321,7 +321,7 @@
 				</div>
 				<div class="scores">
 					<h2>Student Activity</h2>
-					<span id="export_scores_button" class="action_button aux_button" ng-click="exportPopup()">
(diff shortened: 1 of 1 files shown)


fuel/app/themes/default/partials/my_widgets.php
```diff
@@ -321,7 +321,7 @@ class="action_button arrow_down_button">
 				</div>
 				<div class="scores">
 					<h2>Student Activity</h2>
-					<span id="export_scores_button" class="action_button aux_button" ng-click="exportPopup()">
+					<span id="export_scores_button" class="action_button aux_button {{selected.scores.list.length === NULL ? 'disabled' : ''}}" ng-click="selected.scores.list.length === NULL ? angular.noop() : exportPopup()">
```
FrenjaminBanklin: This is a case where https://docs.angularjs.org/api/ng/directive/ngClass would be a bit cleaner to read and wouldn't potentially interfere with the regular `class` application.
```
ng-class="{'disabled': selected.scores.list.length === NULL}"
```

clpetersonucf:
Just needs @FrenjaminBanklin to approve the changes and this should be good to go

Issue #1101, opened 2018-03-06
Hrm. When a user has FULL ACCESS permissions that expire, they can remove them.


iturgeon:
Perhaps if you have FULL ACCESS and have an expiration date, you should not be able to remove it.

Why?  Because a TA may need to be given full access to a widget to be able to alter it's content, but that access should expire.

weining-li:
Should the full access user still be allowed to remove another user's expiration date or change their access level?

clpetersonucf:
Closing as previously addressed and no longer relevant to Materia v11.

PR #1263, opened 2019-10-16
issue/1101 - removed ability for a full access user to remove their own expiration

weining-li:
#1101


clpetersonucf:
Works as intended. I've noticed the grayed-out expiration text is especially difficult to read in Safari:
Screen Shot 2019-11-21 at 2 34 08 PM

I'm still a little concerned about how permission collisions between two "full" access users works though. `User A` grants `User B` FULL access alongside an expiration date. `User B` cannot modify their own expiration date but can potentially remove `User A`'s access. Are we okay with this scenario?

Issue #1112, opened 2018-05-17
Installer Updates


iturgeon:
* Prompt users to ask what domain will serve their assets (we now support cdns for this)
* Fix misspelling of `random private key for encyption` in installer text

iturgeon:
I want to change how configuration is done by default.  Let's use `symfony/dotenv` and use environment variables to do the heavy lifting, it's easier to deal with because we can collect all the variable configuration into one place, it'll work with existing fuelphp config, and it'll be super flexible for all sorts of deployments.

The installer wizard has removed the ability to use the environment config options, I think we should continue to mature that (effectively removing the environment configuration directories from our design - Fuel will still make them (config/migrations.php) but we'll switch to using .env files and the main config files in /fuel/app/config.

We'll need to find an early part of app initiation to load dotenv.


Here's my initial rundown of what the env variables should be:

```
#config.always_load.packages
FUEL_ALWAYS_LOAD_PACKAGES=

#config.always_load.modules
FUEL_ALWAYS_LOAD_MODULES

#config.log_threshold
FUEL_LOG_THRESHOLD

DATABASE_URL=mysql://materia:odin@mysql/materia

#materia.send_emails
SEND_EMAILS=


#materia.urls.static
URLS_STATIC

#materia.urls.engines
# TODO: check if URLS_STATIC can replace this
URLS_ENGINES

#materia.media_quota=5000
MEDIA_QUOTA

#materia.system_email=noReply@materia.com
SYSTEM_EMAIL

#materia.crossdomain=https://static.materia.ucf.edu
# TODO: check if URLS_STATIC SHOULD BE USED HERE
CROSSDOMAIN

#materia.enabled_admin_uploader
ADMIN_UPLOADER_ENABLE

#materia.asset_storage_driver
ASSET_STORAGE_DRIVER

#materia.asset_storage.s3.region
ASSET_STORAGE_S3_REGION


#materia.asset_storage.s3.subdir
ASSET_STORAGE_S3_BASEPATH

#materia.asset_storage.s3.secret_key
ASSET_STORAGE_S3_SECRET

#materia.asset_storage.s3.key
ASSET_STORAGE_S3_KEY

#materia.asset_storage.s3.endpoint
ASSET_STORAGE_S3_ENDPOINT

#materia.asset_storage.s3.bucket
ASSET_STORAGE_S3_BUCKET

#session.driver=db
SESSION_DRIVER=db

#session.memcached.cookie_name
SESSION_NAME

[...]

PR #1297, opened 2020-07-06
Implements dotenv to streamline config  & install

iturgeon:
This reduces the locations of config files that need to be changed down to a single location and enables the use of ENV variables to configure Materia.

#1112


iturgeon:
Re-opened #1300 1300

PR #1299, opened 2020-07-08
[fix] notification email from system user working

iturgeon:
Changes in 7.0.0 changed how object permissions expirations were
handled.  It now purges all old permissions and notifies before checking
for access.  This require notifications to be sent by the server instead
of a specific user. Due to an oversight, lack of testing, and a dev env
that has email disabled - the code that actually sends the mail did not
receive associated updates that it needed.

This solves the problem by extracting the code that resolves the from
user for notifications so it can be used in both places.

This code also changes notifications so that when one is sent, no longer
are all pending notifications sent.  This change was made to prevent old
notifications that have gone unsent from potentially causing errors in
unrelated code (if a bug occurs in sending a notification, it will keep
occuring - blocking other notifications).

This means we no longer have code that will send past failed notifications.

### DIFF
--- a/fuel/app/classes/model/notification.php
+++ b/fuel/app/classes/model/notification.php
@@ -68,24 +68,11 @@ public static function on_widget_delete_event($event_args)
 		switch ($item_type)
 		{
 			case \Materia\Perm::INSTANCE:
-				if ($from_user_id === 0)
-				{
-					// create a mock user
-					$user = (object)[
-						'first' => 'Materia',
-						'last' => '',
-						'username' => 'Server'
-					];
-				}
-				else
-				{
-					$user = \Model_User::find($from_user_id);
-				}
-
+				$from = static::get_user_or_system($from_user_id);
 				$inst = new \Materia\Widget_Instance();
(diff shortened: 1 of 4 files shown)


fuel/app/classes/model/notification.php
```diff
@@ -153,45 +140,57 @@ public static function send_item_notification(int $from_user_id, int $to_user_id
...
-			->get();
+		$from = static::get_user_or_system($this->from_id);
+		$to = \Model_User::find($this->to_id);
+
+		$email->from(\Config::get('materia.system_email'), $from->first.' '.$from->last);
+		$email->reply_to($from->email, $from->first.' '.$from->last);
+		$email->to($to->email, $to->first.' '.$to->last);
+		$email->subject(\Config::get('materia.name').' Notification');
+		$email->html_body($this->subject);
+trace($from->email);
```
clpetersonucf, line 158: Left some trace statements in here by accident?

iturgeon: Gotem

PR #1303, opened 2020-07-20
adds ssl/tls for docker dev

iturgeon:
This should allow for local https development

* changes missing driver config in dev for cache and session
* dramatic change to nginx config to more mimic a production use
* fix for sameSite cookie not gracefully downgrading on non-secure settings

### DIFF
--- a/docker/run_first.sh
+++ b/docker/run_first.sh
@@ -29,6 +29,9 @@ DOCKER_IP="localhost"
 # clean migration files in every environment
 rm -f ../fuel/app/config/**/migrations.php
 
+# generate a self-signed ssl cert
+openssl req -subj '/CN=localhost' -x509 -newkey rsa:4096 -nodes -keyout ./config/nginx/key.pem -out ./config/nginx/cert.pem -days 365
+
 # use env/args to determine which docker-compose files to load
 source run_dc.sh
 
--- a/docker/run_first.sh
+++ b/docker/run_first.sh
@@ -29,6 +29,10 @@ DOCKER_IP="localhost"
 # clean migration files in every environment
 rm -f ../fuel/app/config/**/migrations.php
 
+# make room for the cert (docker may make directories here if `up` was run before the certs exist)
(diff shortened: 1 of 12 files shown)


docker/run_first.sh
```diff
@@ -29,6 +29,9 @@ DOCKER_IP="localhost"
 # clean migration files in every environment
 rm -f ../fuel/app/config/**/migrations.php
 
+# generate a self-signed ssl cert
+openssl req -subj '/CN=localhost' -x509 -newkey rsa:4096 -nodes -keyout ./config/nginx/key.pem -out ./config/nginx/cert.pem -days 365
```
clpetersonucf, line 33: Are there additional requirements to getting this to run successfully? It's failing for me:
```
writing new private key to './config/nginx/key.pem'
./config/nginx/key.pem: Is a directory
```

clpetersonucf: Scratch that, removing `key.pem` and `cert.pem` and rerunning the command above worked. Chalk up another one to "didn't start from scratch and got into a misconfiguration"

clpetersonucf:
Despite starting from absolute scratch, I'm continuing to receive the `invalid certificate authority` warning when trying to visit Materia dev locally

iturgeon:
Yea, it's a self-signed cert, it won't verify as a valid authority.  But this will allow it to work if you either add the generated cert to your lists or allow invalid certs on localhost in your browser

docker/run_first.sh
iturgeon: Hmm, strange.  Not sure what was going on there.

clpetersonucf:
> Yea, it's a self-signed cert, it won't verify as a valid authority. But this will allow https to work if you either add the generated cert to your lists or allow invalid certs on localhost in your browser

Should `run_first.sh` have a post-run message indicating one of those actions is required? Although in hindsight, a self-signed cert being an invalid authority makes sense.

PR #1326, opened 2021-03-16
adds automatic arm support for docker phpfpm, switches to mariadb

iturgeon:
* docker file updated to support arm64 and x64 automatically
* switches docker-compose to use mariadb because mysql doesn't have an arm docker image till v8+


clpetersonucf:

PR #1338, opened 2021-07-28
Hotfix to correct local test failure

clpetersonucf:
The recent docker changes had a side effect of no longer creating an empty test database alongside the primary materia database for the purposes of running unit tests. This re-implements the sql query that's run when the mysql container is created, pending a more thorough solution.


FrenjaminBanklin (approved):
Tests are running, looks like it's working properly.

PR #1342, opened 2021-09-22
Removed browser sync

clpetersonucf:
Nobody uses it and it clogs network and console logs. Removed with extreme prejudice.


FrenjaminBanklin (approved):
Agree.

PR #1344, opened 2021-09-28
Access Request option for LTI previews

clpetersonucf:
Branched off #1341 

Adds a new option to the LTI embed preview screen that's circumstantially enabled if a user is not an owner of the widget. It allows a user to request access to the embedded widget. The selected owner will be given a notification with an option to grant ownership to the person who made the request.

To be paired with https://github.com/ucfopen/Materia-Server-Client-Assets/pull/125

**This pull request includes a migration**

### DIFF
--- a/fuel/app/classes/controller/api/instance.php
+++ b/fuel/app/classes/controller/api/instance.php
@@ -28,4 +28,12 @@ class Controller_Api_Instance extends Controller_Rest
 
 		return $this->response($history, 200);
 	}
+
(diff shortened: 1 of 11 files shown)


FrenjaminBanklin (approved):
This looks good to me.

C/P from the corresponding MSCA PR: The list of users with ownership over a widget in the LTI embed is still pretty spartan, so while I might suggest giving it a little bit of design I don't really have any ideas.

PR #1348, opened 2021-11-08
Dev/9 typhoon

clpetersonucf:
- Widget URLs in the install config now default to `latest` instead of hard-coded, outdated versions.
- Fix for local test failure
- Access Request: a LTI-embedded widget that is accessed by an instructor who does not own a widget now provides a list of widget owners, and the option to request access from any of them. The selected widget owner is sent a notification to provide access to the requesting user
- The `phpcs` git pre-commit hook now defaults to a `full` report instead of `summary`
- browsersync removed, with prejudice
- Added server-side support for `obj` media file types
- Default ASSET_STORAGE_DRIVER in the `docker-compose.yml` is now `file` instead of `s3`
- Added API endpoint to verify instance access permissions
- A big ol' pile of dependabot updates

** NOTE: THIS PR INCLUDES A MIGRATION **


docker/docker-compose.override.test.yml
```diff
@@ -31,8 +31,8 @@ services:
       - MYSQL_USER
       - MYSQL_PASSWORD
       - MYSQL_DATABASE=test
-    volumes:
-      - "./config/mysql/dev.my.cnf:/etc/mysql/conf.d/my.cnf"
+    # volumes:
+    #   - "./config/mysql/dev.my.cnf:/etc/mysql/conf.d/my.cnf"
```
FrenjaminBanklin, line 35: Might prefer just removing this entirely if they're no longer necessary.

fuel/app/classes/materia/widget/asset.php
```diff
@@ -271,6 +273,13 @@ public function render(string $size): void
 			header("Content-Length: {$bytes}");
 			header('Content-Transfer-Encoding: binary');
 			header('Cache-Control: max-age=31536000');
+
+			// Special case for 3D labeling
```
FrenjaminBanklin, line 277: A little more context might be necessary here.

FrenjaminBanklin (approved):
I'm a little wary of always using the 'latest' version of a widget by default. I vaguely remember having this conversation years ago and settling on hard-coding the versions intentionally to avoid some pitfall or another. Maybe everything has stabilized now to the point that we don't need to worry about it.

There are a couple of spots where more clarity in comments would be nice, but otherwise if we're okay with the departure from installing specific widget versions, I see no reason to hold this up.

fuel/app/classes/materia/widget/asset.php
clpetersonucf: Updated comment with additional context.

docker/docker-compose.override.test.yml
clpetersonucf: Removed the lines in question.

ljoks (approved):
Just a bit of a lack of comments for new functions but overall lgtm

PR #1353, opened 2021-12-16
Adds an API endpoint to verify access to a widget instance.

FrenjaminBanklin:
Adds a new API endpoint to verify that the current user has the ability to access a provided widget instance.


clpetersonucf (approved):
Simple and effective. Does the job as required.

PR #1355, opened 2022-01-03
Changes database Docker container from MySQL to MariaDB.

clpetersonucf:
This is a replacement of https://github.com/ucfopen/Materia/pull/1345 which was auto-closed when its base branch was deleted.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -57,7 +57,7 @@ services:
       - fakes3
 
   mysql:
-    image: mysql:5.7.18
+    image: mariadb:10.6.4
     ports:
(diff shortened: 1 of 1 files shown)


FrenjaminBanklin:
Does it still make sense to switch the database engine to MariaDB?

The UCF instance is running in RDS/Aurora on MySQL 5.7, so for our purposes this PR isn't helpful and should probably be closed, but I'm not sure who else might be running Materia or what their constraints may be.

clpetersonucf:
I think you're right. The upgrade path should be from MySQL 5.7 to 8.x, not MariaDB. Closing.

PR #1356, opened 2022-01-03
adds third party oauth check when uploading media files

clpetersonucf:
This is a replacement of https://github.com/ucfopen/Materia/pull/1349 which was auto-closed when its base branch was deleted.


clpetersonucf:

PR #1357, opened 2022-01-03
Issue/1336 add option to hide assets in importer

emanuelGitCodes:


fuel/app/themes/default/partials/catalog/media.php
```diff
@@ -62,6 +61,12 @@
 				</span>
 				<span class='file-date'>
 					{{file.created}}
+					<br>
+					<button ng-class="file.is_deleted == 0 ? 'delete-btn green' : 'delete-btn orange'"
+						ng-click='deleteAsset(file)'>
+						<span ng-if='file.is_deleted == 0'>DELETE</span>
+						<span ng-if='file.is_deleted == 1'>DELETED</span>
```
clpetersonucf, line 68: Given my comments in the review generally, we should probably change the word `DELETED` to be something more directly tied to the action the button will perform in this state - `Restore` or `Undelete` seems more appropriate.

clpetersonucf (changes requested):
This appears to only be half of the implementation we had discussed - the asset is still always visible in the list, with the only difference being the state of the delete/deleted button. Ideally, deleted assets are not visible by default, but I don't see anything in the client-side JS to hide the deleted assets. The behavior should be more like:

If an asset is _not_ deleted: `Delete` button is enabled, asset is visible in the list by default
If an asset _is_ deleted: `Restore` button is enabled, asset is _not_ visible by default, and is only made visible when a separate "Show deleted assets" option is checked.

FrenjaminBanklin:
This one's also covered by #1442, I think?

clpetersonucf:
Yup, this was superseded by and replaced with #1442. Closing.

PR #1362, opened 2022-02-21
adds h5p server to materia stack, along with oauth changes for uploading media

ljoks:
also closes #1356


h5p-server/renderers/player.js
```diff
@@ -0,0 +1,69 @@
...
+
+  function playerInit(contentId = "") {
+    H5P.preventInit = true;
+
+    if (H5PIntegration) {
+      let materiaPath = "";
+      switch (context) {
+        case "prod":
+        case "dev":
+          materiaPath = "https://localhost:8008";
```
clpetersonucf, line 12: This url definitely shouldn't be hard-coded, but rather sourced from an env or other configuration. Doing so usually means pushing an inline variable definition in the fuel controller method that renders this content, via:

```
Js::push_inline('var BASE_URL = "'.Uri::base().'";');
```
Here's an example: https://github.com/ucfopen/Materia/blob/master/fuel/app/classes/controller/widgets.php#L519

h5p-server/express.js
```diff
@@ -0,0 +1,279 @@
...
+
+const setupServer = ([editor, player]) => {
+  h5pEditor = editor;
+  h5pPlayer = player;
+
+  const setupServerPromise = (resolve, reject) => {
+    // TODO replace whitelist with hosts from context env var
+    app.use(
+      cors({
+        origin: [
```
clpetersonucf, line 99: Same as previous comment - these URLs will have to come from a config somewhere.

h5p-server/express.js
```diff
@@ -0,0 +1,279 @@
...
+          console.error("Error GET /play/:type :");
+          console.error(error);
+          res.status(500).end();
+        });
+    });
+
+    // return a new h5p widget
+    // used for materia to check for specific library
+    // should be deprecated as there is no need to save this information on the server
+    app.post("/new/:type", (req, res) => {
```
clpetersonucf, line 249: Is this section actually used? My understanding is we're not actually saving H5P content on the server - which I'm pretty sure was the original intent of the `/new/` post request. I'm not sure if Materia actually performs this request as the comment suggests.

clpetersonucf:
I've left this one open a while as a "well, maybe we'll revisit this", but given H5P's ownership change I don't think there's a chance of moving forward with it. Thanks again for your hard work on this.

ljoks:
Haha appreciate the update, i can assure you im not too heartbroken. I hope Materia is thriving!

PR #1372, opened 2022-04-18
Alpha tags trigger a pre-release instead of a full release

clpetersonucf:
Tested on my own fork and it appears to do the job.


FrenjaminBanklin (approved):
Hard to tell if there are any gotchas in here without seeing it run, but based on the existing action I don't see any reason to believe the new one won't work.

Potential clutter if we release a lot of alpha tags, but that's probably not worth worrying about.

PR #1374, opened 2022-04-20
Security hardening and enhancements

clpetersonucf:
- Fixes inline php tags to prevent server-side code from being rendered client-side.
- Adds HSTS headers to nginx-dev and nginx-prod.

The `X-Frame-Options` header appears to be a non-starter, as `ALLOW-FROM` does not yet have broad browser support and `SAMEORIGIN` disallows requests from the non-static domain.

### DIFF
--- a/docker/config/nginx/nginx-dev.conf
+++ b/docker/config/nginx/nginx-dev.conf
@@ -70,6 +70,9 @@ http {
         listen *:443 ssl;
         listen [::]:443 ssl;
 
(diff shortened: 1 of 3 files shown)


FrenjaminBanklin (approved):
I'm not sure the best way of making sure that the behavior functions as intended re: the Strict-Transport-Security header, but I can at least verify that at least locally, everything still appears to be functioning properly.

PR #1375, opened 2022-04-21
Dev/9.0.1 typhoon hotfixes

clpetersonucf:
- Update references to the widget previously known as Hangman to Guess the Phrase.
- Added some basic CSRF mitigation to v1 API calls.
- The `build_and_release` GitHub action will mark alpha tags as pre-release.
- nginx configurations will now include `remote_addr` in logs by default.
- Bumped the version of mysql slightly to allow support for M1 machines.
- Adjusted permission checks to correct visibility issues for widgets for which a user has read-only access.

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -53,6 +53,7 @@ jobs:
           cd docker && ./run_build_github_release_package.sh ghcr.io/${{ github.repository_owner }}/materia:app-${{ steps.tag_name.outputs.GIT_TAG }}
 
       - name: Upload to Release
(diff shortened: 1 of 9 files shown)


FrenjaminBanklin (approved):
Nothing I can see that necessarily needs work.

MySQL 5.7 is still being supported so the version requirements are fine for now, but at some point we may want to consider moving on to version 8.

I'm also still a little worried that there are edge cases with the CSRF mitigation additions, if not for us then potentially for others down the road. But I don't know enough to predict what they might be.

Looks fine for now, otherwise.

PR #1377, opened 2022-04-25
Added remote_addr to nginx configs

clpetersonucf:


FrenjaminBanklin (approved):
Assuming there aren't any weird nuances to how this works on Heroku, this looks pretty straightforward to me.

PR #1382, opened 2022-05-04
Reworked perm filters for two access level checks

clpetersonucf:
Corrects the two visibility issues related to users with the View Scores permission level.


FrenjaminBanklin (approved):
We might want to audit the perms and prune out the ones that don't actually have any use to prevent future confusion.

This looks good, though.

PR #1390, opened 2022-06-30
Issue/react-cay

cayb0rg:


fuel/app/classes/controller/widgets.php
```diff
@@ -239,101 +240,92 @@ public function get_preview_widget($inst_id, $is_embedded = false)
...
-		$this->theme->set_partial('content', 'partials/widget/no_attempts')
-			->set('classes', 'widget')
-			->set('attempts', $inst->attempts)
-			->set('scores_path', '/scores'.($is_embedded ? '/embed' : '').'/'.$inst->id)
-
-			->set('summary', $this->theme->view('partials/widget/summary')
-				->set('type',$inst->widget->name)
-				->set('name', $inst->name)
-				->set('icon', Config::get('materia.urls.engines')."{$inst->widget->dir}img/icon-92.png"));
+		Js::push_inline('var WIDGET_HEIGHT = "'.$widget->height.'";');
```
clpetersonucf, line 291: `$widget` is undefined here. You probably meant `$inst->widget->height`?

src/components/draft-not-playable.jsx
```diff
@@ -0,0 +1,37 @@
...
+      <section className="page is-draft">
+
+        <h2 className="logo">Sorry, drafts are not playable.</h2>
+
+        <p>You probably need to:</p>
+        <ul>
+          <li>Preview instead of play.</li>
+          <li>Publish this widget to start collecting scores.</li>
+          <li>Check out our documentation.</li>
+          <li>Take a break, watch <a href="http://youtu.be/S1k86fWsmRE?t=35s" style="color:black;">cat</a> <a href="http://www.youtube.com/watch?v=q8Z3NEVIwYg&feature=relmfu" style="color:black;">videos</a>.</li>
```
clpetersonucf, line 20: I actually don't mind that this is here, but the inline style is causing React to choke. With JSX, the style prop requires an object: `style={{ color: 'black' }}`

src/components/scores.jsx
```diff
@@ -0,0 +1,919 @@
...
+
+  let previousAttempts = null
+  if (!hidePreviousAttempts && !isPreview && !guestAccess) {
+    previousAttempts = (
+      <nav className={`header-element previous-attempts ${prevAttemptClass}`} onMouseOver={prevMouseOver} onMouseOut={prevMouseOut}>
+        <h1 onClick={prevMouseOver}>Prev. Attempts</h1>
+        <ul onMouseOver={prevMouseOver}>
+          {attempts.forEach((attempt, index) => (
+            <li key={index}>
+              <a href={`#attempt-${index}`} onClick={attemptClick}>Attempt {index}: <span className="score">{ attempt.roundedPercent}%</span><span className="date">{ dates[index]}</span></a>
```
clpetersonucf, line 689: `dates[index]` is undefined on this line. I think you meant `attemptDates[index]`

src/components/widget-summary.jsx
```diff
@@ -0,0 +1,47 @@
...
+  let bodyRender = null
+  if (!!name) {
+    bodyRender = (
+      <div className="widget_info">
+      	<div className="widget_icon">
+      		<img src={ icon } alt=""/>
+      	</div>
+      	<ul className="widget_info">
+          { name && <li className="widget_name">{ name }</li>}
+          { avail && <li className="widget_availability">{ name }</li>}
```
clpetersonucf, line 34: `avail` is undefined here - in the original partial view, it's exposed to the view from the `build_widget_login` method of the widgets controller. It'll probably have to be added as a JS global.

clpetersonucf (changes requested):
I've been working on actually viewing each of the new pages/components in context in the application, and am running into a few cases where values are missing or incorrect. Understandably a lot of these are edge case views to begin with. The requested changes here probably aren't comprehensive yet - I'll add more feedback as I go.

src/util/api.js
```diff
@@ -387,11 +461,16 @@ export const apiGetPlaySession = ({ widgetId }) => {
...
 }
 
-export const apiGetQuestionSet = (instId, playId) => {
+export const apiGetQuestionSet = (instId, playId = null) => {
 	return fetch('/api/json/question_set_get/', fetchOptions({ body: `data=${formatFetchBody([instId, playId])}` }))
 		.then(qset => qset.json())
 }
 
+export const apiGetQuestionSetHistory = (instId) => {
+	return fetch('/api/instance/history', fetchOptions({ body: `data=${formatFetchBody([instId])}` }))
```
clpetersonucf, line 470: I can understand the confusion with the API calls, but this is using the formatting required for the old (v1) API: those hitting `/api/json/xxx`. Newer API endpoints are added within the requisite controllers, in this case `controllers/api/instance.php`. This particular request is formatted for v1, which requires everything to be POST - but the instance API controller is expecting this particular call to be a GET request. It currently 405s.

fuel/app/classes/controller/widgets.php
cayb0rg: Oh, yep. The WIDGET_HEIGHT variable isn't being used, though, so I'll just remove it.

PR #1391, opened 2022-07-19
Merge Changes for Media Importer

emanuelGitCodes:
Changes included
Adding migration for the deletion of assets in the front end.


fuel/app/classes/controller/media.php
```diff
@@ -13,15 +17,16 @@ class Controller_Media extends Controller
 
 	// overrides Trait_CommonControllerTemplate->before()
 	public function before()
```
clpetersonucf, line 19: If the controller's `before` method is empty, we can remove it completely.

fuel/app/classes/controller/media.php
```diff
@@ -38,26 +43,49 @@ public function get_render($asset_id, $size='original')
...
+				->set('title', 'Media Catalog')
+				->set('page_type', 'import');
+
+			$theme->set_partial('footer', 'partials/angular_alert');
+			$theme->set_partial('content', 'partials/catalog/media');
+
+			return Response::forge($theme->render());
+		}
+	*/
+
```
clpetersonucf, line 69: Remove commented code if it isn't required anymore (I still do this too - no worries)

fuel/app/classes/controller/media.php
```diff
@@ -38,26 +43,49 @@ public function get_render($asset_id, $size='original')
...
 			->set('page_type', 'import');
 
-		$theme->set_partial('footer', 'partials/angular_alert');
-		$theme->set_partial('content', 'partials/catalog/media');
+		Css::push_group(['media_import']);
+		Js::push_group(['react', 'media']);
 
-		return Response::forge($theme->render());
+		// Forge is needed to render content of the iframe.
+		return Response::forge($this->theme->render());
```
clpetersonucf, line 88: I would double-check if this is the case - I think the `forge()` method can be removed completely with the transition to react. If you look at other controller methods, `forge()` isn't referenced.

fuel/app/config/config.php
```diff
@@ -101,11 +101,9 @@
 	'log_date_format'  => 'H:i:s',
 
 	// provide a monolog handler
-	'log_handler_factory' => function($locals, $level)
-	{
+	'log_handler_factory' => function ($locals, $level) {
```
clpetersonucf, line 104: Do you have a formatter or linter that's moving the braces here? The `phpcs` linter, which is run during the pre-commit checks, vocally prefers braces on the next line. This change (and the one on line 106) shouldn't be necessary.

fuel/app/config/config.php
```diff
@@ -333,8 +331,8 @@
 		],
 		[
 			'id' => 2,
-			'package'  => 'https://github.com/ucfopen/hangman-materia-widget/releases/latest/download/hangman.wigt',
-			'checksum' => 'https://github.com/ucfopen/hangman-materia-widget/releases/latest/download/hangman-build-info.yml',
+			'package'  => 'https://github.com/ucfopen/guess-the-phrase-materia-widget/releases/latest/download/guess-the-phrase.wigt',
+			'checksum' => 'https://github.com/ucfopen/guess-the-phrase-materia-widget/releases/latest/download/guess-the-phrase-build-info.yml',
```
clpetersonucf, line 335: These changes should be redundant at this point - this change should already be reflected on the main react branch.

clpetersonucf (changes requested):
In addition to the inline code comments I left, a few higher-level things:
- The changes to `fuel/app/config/config.php` seem unnecessary - it's just small formatting changes. They may be worthwhile to include, but I'd rather not bundle them with this particular pull request - since they're not really in the scope of the issue being added.
- The changes to `fuel/app/config/js.php` (adding spaces on either side of the concatenator) are not necessary.
- Modifying `fuel/app/themes/default/partials/catalog/media.php` is redundant. These are the old PHP partial views that are being retired in favor of the react pages that will be rendered instead.

clpetersonucf:
Also, I would be sure to resolve the merge conflict that currently exists on this branch. You can resolve it by merging in the base branch (`issue/support-dashboard-in-react`), handling the individual conflicts, and committing the changes, then pushing the new commit(s) to the remote.

clpetersonucf:
I've stripped out the substantive changes from here (minus stuff prettier added) and will be carrying this over the finish line in #1442

PR #1397, opened 2022-09-01
Issue/my widgets pagination

emanuelGitCodes:
Add pagination to the widgets instance when loading the page.

### DIFF
--- a/src/components/my-widgets-page.jsx
+++ b/src/components/my-widgets-page.jsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect, useMemo } from 'react'
 import { useQuery } from 'react-query'
-import { apiGetWidgetInstances, apiGetUser, readFromStorage, apiGetUserPermsForInstance } from '../util/api'
+import { apiGetWidgetInstances, apiGetPaginatedWidgetInstances, apiGetUser, readFromStorage, apiGetUserPermsForInstance } from '../util/api'
 import rawPermsToObj from '../util/raw-perms-to-object'
 import Header from './header'
 import MyWidgetsSideBar from './my-widgets-side-bar'
@@ -25,7 +25,7 @@ const randomBeard = () => {
 }
 
 const initState = () => {
-	return({
+	return ({
 		selectedInst: null,
 		otherUserPerms: null,
 		myPerms: null,
@@ -46,17 +46,22 @@ const MyWidgetsPage = () => {
 	const copyWidget = useCopyWidget()
 	const deleteWidget = useDeleteWidget()
 	readFromStorage()
-	const { data: widgets, isLoading, isFetching } = useQuery({
-		queryKey: 'widgets',
-		queryFn: apiGetWidgetInstances,
-		staleTime: Infinity
-	})
-	const { data: user} = useQuery({
+
+	const [page, setPage] = useState(1)
+	const [widgetsList, setWidgetsList] = useState([])
+	const {
+		data,
+		isLoading,
+		isSuccess,
+		isFetching,
+	} = useQuery('widgets', () => apiGetPaginatedWidgetInstances(page), { keepPreviousData: true })
+
+	const { data: user } = useQuery({
 		queryKey: 'user',
 		queryFn: apiGetUser,
 		staleTime: Infinity
 	})
-	const { data: permUsers} = useQuery({
+	const { data: permUsers } = useQuery({
 		queryKey: ['user-perms', state.selectedInst?.id],
 		queryFn: () => apiGetUserPermsForInstance(state.selectedInst?.id),
 		enabled: !!state.selectedInst && !!state.selectedInst.id && state.selectedInst?.id !== undefined,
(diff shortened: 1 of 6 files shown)


src/components/my-widgets-page.jsx
```diff
@@ -46,17 +46,22 @@ const MyWidgetsPage = () => {
...
-	const { data: user} = useQuery({
+
+	const [page, setPage] = useState(1)
+	const [widgetsList, setWidgetsList] = useState([])
+	const {
+		data,
+		isLoading,
+		isFetching,
+		refetch,
+	} = useQuery('widgets', () => apiGetPaginatedWidgetInstances(page), { keepPreviousData: true, })
```
clpetersonucf, line 57: One thing I'm noticing is that this query will continue to ping the server periodically with a page value above the `total_num_pages` value returned by the server response. Ideally, once all pages are fetched, the query no longer needs to run continually. There are a couple ways to disable the query, like adding a condition to the `enabled` option, or the `refetch` or `retry` options.

src/components/my-widgets-page.jsx
```diff
@@ -75,30 +80,40 @@ const MyWidgetsPage = () => {
...
 	// isFetching - actual data, from API
 	useEffect(() => {
 		if (!isFetching) {
-			checkPreselectedWidgetAccess(widgets)
+			if (page <= data.total_num_pages) {
+				setPage(page + 1)
+			}
+			setWidgetsList(current => [...current, ...data.pagination])
 		}
 	}, [isFetching])
```
clpetersonucf, line 103: Attaching a `useEffect` to `isFetching` is a little unconventional, but since that useEffect was already being used, we may as well roll with it. In addition to my other suggestion for disabling the query once all pages are fetched, I think it might be useful to assign `isFetching` to a value with a name that makes it easier to identify. You can see examples of this on lines 59 and 64, something like:
```
const { data: instancesFetching } = useQuery(...)
```

Obviously the references to `isFetching` on lines 97 and 103 would have to be updated to the new value.

src/components/my-widgets-page.jsx
```diff
@@ -123,29 +138,29 @@ const MyWidgetsPage = () => {
...
-		setState({...state, selectedInst: inst, noAccess: false, currentBeard: beards[index], postFetch: false})
+		setState({ ...state, selectedInst: inst, noAccess: false, currentBeard: beards[index], postFetch: false })
 		setUrl(inst)
 	}
 
 	const onCopy = (instId, newTitle, newPerm, inst) => {
-		setState({...state, selectedInst: null})
-
+		setState({ ...state, selectedInst: null })
+		console.log(inst.widget)
```
clpetersonucf, line 163: Make sure to remove this console.log

src/util/api.js
```diff
@@ -43,6 +43,19 @@ export const apiGetWidgetInstances = () => {
 		})
 }
 
+export const apiGetPaginatedWidgetInstances = page_number => {
```
clpetersonucf, line 46: This function appears to do the job, it might be worth it to see if the `apiGetWidgetInstances` function is being used by anything else in Materia's front end, and if not, just replace `apiGetWidgetInstances` with the new version (but keeping the `apiGetWidgetInstances` name). Having both a paginated and non-paginated version would be redundant.

clpetersonucf (changes requested):
Just some small changes being requested here. I also don't see the other paginated query you added - for play scores - present here, make sure those changes are pushed to your remote so they get included in this PR.

src/util/api.js
emanuelGitCodes: The API is not used anywhere else, so I'll make that update.

src/components/my-widgets-page.jsx
emanuelGitCodes: I solved this by implementing `refetchOnWindowFocus: false` to the query options.

emanuelGitCodes:
A bunch of shenanigans happened with the file `widget-creator.jsx`
Updated it to be the same as the one in the `support react branch`

clpetersonucf (changes requested):
It looks like the pagination behavior is fixed and consistent and it seems to work well. In addition to the missing style comment I left below, some other problems I'm seeing:
- Loading the my widgets page with a widget hash pre-selected (ex: `127.0.0.1/my-widgets#pOixs`) no longer selects the instance once the page has loaded, but rather says I don't have access to the widget. If I select another instance from the list on the left, it loads normally.
- Copying widget instances appears to be broken.
- Deleting widget instances appears to be broken.
- The list of instances on the side bar no longer appears to be listed in descending order based on creation date. I made a bunch of copies of a widget (on the default react branch), and when I switch to the pagination branch, the list of instances are randomly ordered. It's not completely random; I'm wondering if the list is being created based on the order individual pages are returned from the server.

emanuelGitCodes:
I'll look into the first 3 points to see what's happening.
For the final point, I assume the API call is organizing the widgets, but when they get to the frontend, the other is changed once they are appended to the pagination array.

emanuelGitCodes:
Looking over the code, I found a possible bug that my code introduces into the project.
It has to do with the files `useDeleteWidget.jsx` and `useCopyWidget.jsx`.
I'll look into it and return with a response asap.

clpetersonucf (changes requested):
After additional testing, I was ready to call this one good - but after merging in upstream changes from the main react branch, I'm noticing that copying widgets is no longer updating the selected widget and the widget list appropriately. Curiously, if I make a copy of the copied widget, the UI works as it should. Once that issue is resolved, I'm ready to call it good and merge it in.

emanuelGitCodes:
When creating the component, I copied the main react branch, and from there, I started development. Maybe something new merged into the main react branch, causing the component to break. Either way, I'll be looking into it.

emanuelGitCodes:
Look, we don't have to talk about it. The rebase never happened.

clpetersonucf (approved):
Since the current state of the branch is functional, I'm giving this branch tentative approval. We should still make sure it remains functional with the current head of the react branch.

clpetersonucf:
This feels pretty robust now. I'm happy with it. Merging.

PR #1398, opened 2022-09-06
Issue/reactify widget admin

cayb0rg:
Converts the Widget Admin page to React.


src/components/widget-list-card.jsx
```diff
@@ -0,0 +1,259 @@
...
+        questionTypes = state.widget.meta_data.supported_data.map((qtype, i) => <li>{ qtype }</li>)
+    }
+
+    let exportOptions = null
+    if (state.widget.meta_data.playdata_exporters) {
+        exportOptions = state.widget.meta_data.playdata_exporters.map((qtype, i) => <li>{ qtype }</li>)
+    }
+
+    return (
+        <li>
```
clpetersonucf, line 131: Even though these aren't being iteratively generated as part of a map function or anything, react is still complaining about the lack of a `key` prop. I'd see about adding one to this list element.

src/components/widget-list-card.jsx
```diff
@@ -0,0 +1,259 @@
...
+                        </span>
+                    </div>
+                    <div>
+                        <span>
+                            <label>Settings:</label>
+                        </span>
+                        <span>
+                            <div>
+                                <label className="normal">
+                                    <input type="checkbox" defaultChecked={state.widget.in_catalog} checked={state.widget.in_catalog}
```
clpetersonucf, line 165: React also really doesn't like inputs having both a `defaultChecked` prop and a `checked` prop simultaneously - I think I'd err on the side of keeping `checked={state.widget.in_catalog}` and just ensuring that the state value is set properly when the component is mounted.

src/components/widget-list-card.jsx
```diff
@@ -0,0 +1,259 @@
...
+                                    onClick={toggleInCatalog}/>
+                                    In Catalog
+                                </label>
+                            </div>
+                            <div>
+                                <label className="normal">
+                                    <input type="checkbox" defaultChecked={state.widget.is_editable} checked={state.widget.is_editable}
+                                    onClick={toggleIsEditable}/>
```
clpetersonucf, line 173: In addition to the feedback above, I think these inputs should be using `onChange` instead of `onClick` - even if it appears to work just fine as it is now.

clpetersonucf (changes requested):
Looks good and works just fine in testing. I provided some small feedback items based on errors React was throwing in the console. Once those are fixed, it's good to go.

clpetersonucf (approved):
LGTM :+1:

Issue #1403, opened 2022-10-11
[React branch] evaluate player and creator components for race conditions


clpetersonucf:
The Materia side of the creator and player pull data from two sources: asynchronous calls to the server, and postMessages from the widget iframe. The mixture can cause race conditions, such as in the creator component's `onCreatorReady` function. We should take another look at both the creator and player and ensure all other race conditions due to asynchronous queries and postMessages are appropriately handled.

clpetersonucf:
These appear to be resolved.

PR #1406, opened 2022-10-18
Issue #1403/#1404 - a take on race conditions in creator and player

cayb0rg:


src/components/widget-creator.jsx
```diff
@@ -107,100 +120,93 @@ const WidgetCreator = ({instId, widgetId, minHeight='', minWidth=''}) => {
...
+			if (widgetInstance && (!keepQSet || Object.keys(keepQSet).length == 0)) {
+				return;
+			} else if (widgetInstance && frameRef.current == null) {
+				onInitFail('Unable to load widget.')
+			} else if (widgetInstance && !canPublish && !widgetInstance.is_draft) {
+				onInitFail('Widget type can not be edited by students after publishing.')
 			} else {
-				embedCreator();
-			}
+				if (widgetInstance && (keepQSet || qsetToReload)) {
```
clpetersonucf, line 153: Just with functional testing the changes, I'm noticing that there's still an occasional hiccup where this condition is reached and resolves as false even for existing widgets. As best I can tell, since this entire check is wrapped in a `useEffect` that's set to fire on both changes to `creatorReady` and `keepQSet`, there are some situations where `keepQSet` has not populated when `creatorReady` is set to `true`, and when it finally updates, `creatorReady` has flipped back to false. I'm not 100% sure on a resolution other than changing when and how `setCreatorReady(false)` is called.

clpetersonucf (changes requested):

PR #1412, opened 2022-10-31
dev/9.0.2

clpetersonucf:
This is another small minor patch prior to the big 10.0 update to address the upcoming end-of-life for PHP 7.4. @iturgeon has graciously contributed the required updates to bump the system to PHP 8.1 support. Prior to release, we should probably include MySQL updates as well.


FrenjaminBanklin (approved):
Everything I could think to check when testing this both locally and in the field seems to be working properly.

PR #1415, opened 2022-11-08
Converts LTI views to React

cayb0rg:
Changes
- Converts LTI views to React
- Removes jquery from test_provider and test_learner pages
- Updates deprecated assertContains to assertContainsEquals in test functions


clpetersonucf (changes requested):
Some high level design comments:

- It looks like you're trying to re-implement `get_owners` and `request_access` in the v1 API, but those API requests actually already exist - they're just not in v1. I know, it's confusing. `post_request_access` is a method in `fuel/app/classes/controller/api/instance.php`. The API in Materia is currently not great: older API requests are in v1, and newer API requests are in their respective API controller. It's something that has to be rebuilt from the ground up in the future python rewrite. `get_owners` probably isn't even required, the ownership is assessed in the LTI preview controller, and we're better off emitting that flag to the window like other JS constants in other React views.
- Generally, try and avoid making changes to any of the old (or creating new) stylesheets in `src/css`. My goal is to kill these completely, and implement CSS in one of two places: a common stylesheet (`include.scss`) in the components directory, or a component-specific stylesheet in that same directory. With several LTI components making use of similar styles, it'd probably make sense for them all to inherit the same stylesheet.
- I noticed you made a change to the `post_login.php` partial under `themes/default/lti/partials`. That's also part of the deprecated file structure, as those views are all being removed. Anything you're changing there should be changed in the corresponding components instead. The individual React views and their components are replacing the old PHP partials.

clpetersonucf:
Updated with changes to the main react branch. Something I'm noticing: the LTI picker isn't loading any widgets, which I suspect is due to pagination in the `widget_paginate_instances_get`, but haven't looked at it in detail.

clpetersonucf (changes requested):
The diff is looking better, I don't see any glaring issues.

The test suite is failing consistently:
```
There was 1 failure:

1) Test_Api_V1::test_widget_instance_lock_for_another_user
Failed asserting that false is true.

/var/www/html/fuel/app/tests/api/v1.php:496
```

The only other thing I'm noticing is that attempting to test LTI Assignment Launch in the test provider yields the message `LTI Assignment URL can not be blank!` even when the the assignment URL is populated.

fuel/app/classes/materia/api/v1.php
```diff
@@ -1185,4 +1185,28 @@ static private function _get_instance_for_play_id($play_id)
 		if ( ! ($inst = Widget_Instance_Manager::get($inst_id))) throw new \HttpNotFoundException;
 		return $inst;
 	}
+
+	// Sends a notification to user requesting access to instance
+	static public function request_access(string $inst_id, int $owner_id)
```
clpetersonucf, line 1190: This API call is unnecessary, newer API endpoints are in the API controllers. `post_request_access` is in `fuel/app/classes/controller/api/instance.php`.

clpetersonucf:
> The only other thing I'm noticing is that attempting to test LTI Assignment Launch in the test provider yields the message `LTI Assignment URL can not be blank!` even when the the assignment URL is populated.

Update: this is probably because the test provider was relying on jquery for basic functionality, which is no longer bundled with Materia. We can probably update the page to use a CDN link for jQuery. I'd rather not bundle it with Materia's JS and CSS assets.

fuel/app/modules/lti/classes/controller/lti.php
```diff
@@ -116,29 +107,34 @@ public function action_success(string $inst_id)
...
-			->set('widget_name', $inst->widget->name)
-			->set('preview_url', \Uri::create('/preview/'.$inst_id))
-			->set('icon', \Config::get('materia.urls.engines')."{$inst->widget->dir}img/icon-92.png")
-			->set('preview_embed_url', \Uri::create('/preview-embed/'.$inst_id))
-			->set('current_user_owns', $current_user_owns)
-			->set('instance_owner_list', $instance_owner_list);
-
-		$this->insert_analytics();
+		$instance_owner_list = $current_user_owns ? [] : (array_map(function ($object)
+		{
```
FrenjaminBanklin, line 112: Try to clean up any trailing whitespace.

fuel/app/modules/lti/classes/controller/lti.php
```diff
@@ -83,28 +79,23 @@ public function action_picker(bool $authenticate = true)
...
 
+		$this->theme->set_template('layouts/react');
 		$this->theme->get_template()
 			->set('title', 'Select a Widget for Use in '.$system)
 			->set('page_type', 'lti-select');
-
-		$this->theme->set_partial('content', 'partials/select_item');
-		$this->theme->set_partial('header', 'partials/header_empty');
-		$this->insert_analytics();
+
```
FrenjaminBanklin, line 97: Try to clean up any trailing whitespace.

fuel/app/modules/lti/classes/controller/lti.php
```diff
@@ -116,29 +107,34 @@ public function action_success(string $inst_id)
...
+		\Js::push_inline('var OWNER_LIST = '.json_encode($instance_owner_list).';');
+		\Js::push_inline('var USER_ID = "'.\Model_User::find_current_id().'";');
 
 		\Css::push_group(['core', 'lti']);
 
+		$this->theme->set_template('layouts/react');
+		$this->theme->get_template()
+			->set('title', 'Widget Connected Successfully')
+			->set('page_type', 'preview');
+
```
FrenjaminBanklin, line 135: Try to clean up any trailing whitespace.

fuel/app/tests/api/v1.php
```diff
@@ -1564,4 +1569,4 @@ protected function assert_validation_error_message($msg)
 		$this->assertInstanceOf('\Materia\Msg', $msg);
 		$this->assertEquals('Validation Error', $msg->title);
 	}
-}
+}
```
FrenjaminBanklin, line 1572: Trailing new line is good practice.

fuel/app/themes/default/lti/partials/config_xml.php
FrenjaminBanklin, line 1: This file is still being loaded when visiting `https://site_url/lti/`.

Is this being replaced with something? If not, restore this file.

fuel/app/modules/lti/classes/controller/error.php
FrenjaminBanklin, line 1: It looks like the specific LTI error partials have been removed in favor of putting all errors through a single function and controlling what shows up on the page that way?

If so, clean up unnecessary variable usage in the relevant controllers and consider adding additional buttons to the LTI test provider to make sure all of the cases are covered.

fuel/app/themes/default/lti/partials/error_autoplay_misconfigured.php
FrenjaminBanklin, line 1: See above message re: consolidating LTI errors. Make sure there's something equivalent to this removed partial in the new code, and make sure there's a way for us to test it.

fuel/app/themes/default/lti/partials/error_lti_guest_mode.php
FrenjaminBanklin, line 1: See above message re: consolidating LTI errors. Make sure there's something equivalent to this removed partial in the new code, and make sure there's a way for us to test it.

fuel/app/themes/default/lti/partials/error_unknown_assignment.php
FrenjaminBanklin, line 1: See above message re: consolidating LTI errors. Make sure there's something equivalent to this removed partial in the new code, and make sure there's a way for us to test it.

fuel/app/themes/default/lti/partials/error_unknown_user.php
FrenjaminBanklin, line 1: See above message re: consolidating LTI errors. Make sure there's something equivalent to this removed partial in the new code, and make sure there's a way for us to test it.

src/components/lti/select-item.jsx
```diff
@@ -0,0 +1,223 @@
...
+                <a role="button" id="create-widget-button" onClick={() => setShowRefreshArrow(true)} className="button" target="_blank" href={window.BASE_URL + "/widgets"}>Create a widget at Materia</a>
+            </div>
+        </div>
+    } else {
+        createNewInstanceLink = <a id="goto-new-widgets" onClick={() => setShowRefreshArrow(true)} className="external" target="_blank" href={window.BASE_URL + "widgets"}>Or, create a new widget at Materia</a>
+    }
+
+    let sectionRender = null
+    if (displayState == 'selectInstance') {
+        sectionRender =
```
FrenjaminBanklin, line 179: Try to clean up trailing whitespace.

src/components/lti/select-item.jsx
```diff
@@ -0,0 +1,223 @@
...
+            
+            if (!!window.RETURN_URL) {
+                // add a ? or & depending on window.RETURN_URL already containing query params
+                const separator = window.RETURN_URL.includes('?') ? '&' : '?'
+                // encode the url
+                const url = encodeURI(selectedInstance.embed_url)
+                // redirect the client to the return url with our new variables
+                window.location = `${window.RETURN_URL}${separator}embed_type=basic_lti&url=${url}`
+            }
+        }
```
FrenjaminBanklin, line 94: Try to clean up trailing whitespace.

src/components/lti/select-item.jsx
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    useEffect(() => {
+        // End progress bar
+        if (progressComplete && !!selectedInstance) {
+            let pg = document.querySelector('.progress-container')
+            let pgSpan = document.querySelector('.progress-container span')
+            pg.classList.add('success')
+            pgSpan.innerText = 'Success!'
+
```
FrenjaminBanklin, line 85: Try to clean up trailing whitespace.

src/components/lti/select-item.jsx
```diff
@@ -0,0 +1,223 @@
...
+    const [state, setState] = useState({
+		page: 1,
+        instances: [],
+    })
+
+    const { data, isFetching: isFetching, refetch: refetchInstances} = useQuery({
+		queryKey: 'instances',
+		queryFn: () => apiGetWidgetInstances(state.page),
+        staleTime: Infinity,
```
FrenjaminBanklin, line 24: Try to clean up trailing whitespace.

fuel/app/themes/default/lti/partials/outcomes_xml.php
FrenjaminBanklin, line 1: This seems like it's still being loaded as part of the `on_score_updated_event` LTI event. Is it being replaced with something? If not, restore this file.

FrenjaminBanklin (changes requested):
It looks like the LTI-relevant XML files got caught up in sweeping deletions, but those should probably be restored.

There are a few spots I noticed that could use some cleanup re: extraneous white space and removing anything left over from debugging.

The LTI test provider's behavior seems a bit unpredictable, but I haven't really been able to nail down a pattern. Buttons that should be sending requests with bad OAuth signatures appear not to be causing any problems when they probably should, for example. Since so much is being changed, this may also be a good time to add the means to make additional checks - loading a widget as an instructor who does not own it, loading a widget in guest mode, loading an unknown assignment, etc.

Also, though it's minor, it looks like the course navigation screen's appearance is different. The two columns should be taking up the same amount of vertical space, but one is taller than the other.

clpetersonucf (approved):

PR #1428, opened 2022-12-28
Sakai LTI compatibility. LTI module combined with app.

### LINKED ISSUE #1418: If the referrer url is longer than 255 characters, creating a new play will fail
I discovered this due to a long referrer url filled with a bunch of get params.  When the play is created, it looks up the referrer and inserts it into the database.  If it's extra long (longer than the db column limits), the insert will fail. 

Suggest just truncating the referrer to 255, though increasing the length of the column is a possibility too.

iturgeon:
This is a work in progress. 

Add LTI compatibility with The [Sakai LMS](https://www.sakailms.org/)

* /lti/assignment will now accept `ContentItemSelectionRequest` launches
* Change var to const for js constants being injected into the page from the server
* Allows the before_play_start event to determine if the lti content item picker should be displayed
* Move picker display function into a common controller
* Add an api method that’ll validate and sign a content item selection lti request
* injects LTI_MESSAGE_TYPE and LTI_KEY js constants into the picker page template
* LTILaunch gains a method to load the lti config that is associated with a lti key
* Updated a few strings that specifically mention canvas
* Moves all classes, tests, migrations, and configs from fuel/app/modules/lti into fuel/app
* simplify lti launch picker code when sending content selection messages to lti/assignment
* fixed issue with starting plays with referrer urls longer than 255 characters fixes https://github.com/ucfopen/Materia/issues/1418
* Overhaul lit test provider a bit - mostly to simplify
* attempt to fix yarn cache issue
* Changes docker build process to ensure the built image has admin:make_paths_writable run to avoid having to run it every startup
* new api endpoint for signing content item selection `lti_sign_content_item_selection`
* improves oauth validation in fuel by ensuring the request uri is not masked by FuelPHP

Some specific testing should take place before merging:
* [ ] Test Obojobo Next compatibility, it may not support the new picker integration
* [ ] Test Canvas integration. (I tested the picker in an assignment on Canvas Free for Teachers)

This work is sponsored by [LongSight](https://www.longsight.com/)


docker/run_build_assets.sh
```diff
@@ -22,4 +22,4 @@ docker run \
 	--mount type=bind,source="$(pwd)"/../,target=/build \
 	--mount source=materia-asset-build-vol,target=/build/node_modules \
 	node:12.11.1-alpine \
-	/bin/ash -c "apk add --no-cache git && cd build && yarn install --frozen-lockfile --non-interactive --production --silent --pure-lockfile --force"
```
iturgeon, line 25: My Docker build process was dying regularly for some reason, this seems to have helped.

fuel/app/classes/ltilaunch.php
```diff
@@ -53,7 +52,7 @@ public static function from_request()
 		return static::$launch;
 	}
 
-	public static function config()
+	public static function config_from_request()
```
iturgeon, line 55: Renamed because i added `config_from_key`

fuel/app/classes/ltilaunch.php
```diff
@@ -77,6 +76,23 @@ public static function config()
 		return static::$config;
 	}
 
+	public static function config_from_key($key)
```
iturgeon, line 79: Needed for cases where the request isn't a regular lti launch - for instance, when the frontend loads a signed request from the backend

fuel/app/classes/materia/session/play.php
```diff
@@ -65,7 +65,7 @@ public function start($user_id=0, $inst_id=0, $context_id=false, $is_preview=fal
 			// Essentially true but fragile.
 			$is_lti = array_key_exists('lti_message_type', $this->environment_data['input']) || array_key_exists('token', $this->environment_data['input']);
 			$this->auth = $is_lti ? 'lti' : '';
-			$this->referrer_url = \Input::referrer();
+			$this->referrer_url = mb_strimwidth(\Input::referrer(), 0, 255);
```
iturgeon, line 68: The referrer url can be longer than the database will handle, this will limit it.

fuel/app/classes/oauth.php
```diff
@@ -10,24 +9,24 @@ public static function validate_post()
...
 			if (empty($signature)) throw new \Exception('Authorization signature is missing.');
 			if (empty($nonce)) throw new \Exception('Authorization fingerprint is missing.');
-			if (\Input::post('oauth_consumer_key') !== $lti_config['key']) throw new \Exception('Authorization signature failure.');
+			if (\Input::post('oauth_consumer_key') !== $lti_config['key']) throw new \Exception('Authorization key failure.');
 			if ($timestamp < (time() - $lti_config['timeout'])) throw new \Exception('Authorization signature is too old.');
 
 			$hasher   = new \Eher\OAuth\HmacSha1(); // THIS CODE ASSUMES HMACSHA1, could be more versetile, but hey
 			$consumer = new \Eher\OAuth\Consumer(null, $lti_config['secret']);
-			$request  = \Eher\OAuth\Request::from_consumer_and_token($consumer, null, 'POST', \Uri::current(), \Input::post());
+			$request  = \Eher\OAuth\Request::from_consumer_and_token($consumer, null, 'POST', \Uri::main(), \Input::post());
```
iturgeon, line 21: main() is more accurate for the purpose of signing - FuelPHP can internally re-route to another uri - signing requires validating off the original url the server received.

fuel/app/config/lti.php
```diff
@@ -1,6 +1,115 @@
...
+			'upgrade_to_launch_url' => true,
+
+			// Security Settings CHANGE SECRET AT LEAST!!!
+			'secret'                => $_ENV['LTI_SECRET'],
+			'key'                   => $_ENV['LTI_KEY'],
+
+			// When sending LTI ContentItem Selection requests, Sakai updated and required
+			// that GET params in the url aren't duplicated in the body of params
+			// as they caused oauth signature validation to fail in Sakai.
+			'enforce_unique_params' => $_ENV['LTI_ENFORCE_UNIQUE_PARAMS'] ?? true,
```
iturgeon, line 85: I added this option as it's not super clear if other platforms will want this or not.

fuel/app/config/routes.php
```diff
@@ -34,7 +34,14 @@
 	'preview/(:alnum)(/.*)?'               => 'widgets/preview_widget/$1',
 	'preview-embed/(:alnum)(/.*)?'         => 'widgets/play_embedded_preview/$1',
 	'embed/(:alnum)(/.*)?'                 => 'widgets/play_embedded/$1',
-	'lti/assignment?'                      => 'widgets/play_embedded/$1', // legacy LTI url
+	'lti/assignment?' => function () {
```
iturgeon, line 37: We don't use this much in Materia, so this may be difficult to find in the future.  BUT, Fuel lets us add a function here to determine what controller receives a request.  This is used to support lti implementations that only send requests to a single Materia url.  To support this, lti/assignment will allow picker or content selection requests as well as widget play requests.

materia-app.Dockerfile
```diff
@@ -65,15 +65,15 @@ RUN composer install --no-cache --no-dev --no-progress --no-scripts --prefer-dis
 # =====================================================================================================
 FROM node:12.11.1-alpine AS yarn_stage
 
-RUN apk add --no-cache git
+RUN apk add --no-cache git python make g++
```
iturgeon, line 68: I ran into issues on new base php images building that were helped by adding these

materia-app.Dockerfile
```diff
@@ -89,3 +89,15 @@ USER www-data
 COPY --from=composer_stage --chown=www-data:www-data /var/www/html /var/www/html
 COPY --from=yarn_stage --chown=www-data:www-data /build/public /var/www/html/public
 COPY --from=yarn_stage --chown=www-data:www-data /build/fuel/app/config/asset_hash.json /var/www/html/fuel/app/config/asset_hash.json
+
+# set bogus values so admin task can run
+ENV AUTH_SIMPLEAUTH_SALT=TEMPINVALIDVALUE
```
iturgeon, line 94: php8 was unhappy running admin commands without these environment vars

package.json
```diff
@@ -19,7 +19,7 @@
   },
   "dependencies": {
     "fs-extra": "^8.0.1",
-    "materia-server-client-assets": "2.4.2",
+    "materia-server-client-assets": "https://github.com/iturgeon/Materia-Server-Client-Assets.git#842eaadcf38811bbaa37d804049cc36a4f643a9b",
```
iturgeon, line 22: THIS NEEDS TO BE UPDATED WHEN supporting MSCA PR is merged/tagged

clpetersonucf:
An update on this: I plan on addressing and implementing the changes on this PR after the react branch has shipped (which is hopefully imminent). The major changes required will be converting the PHP partials into their corresponding react views, and integrating the related changes to MSCA, since that repo is being deprecated.

iturgeon:
> An update on this: I plan on addressing and implementing the changes on this PR after the react branch has shipped (which is hopefully imminent). The major changes required will be converting the PHP partials into their corresponding react views, and integrating the related changes to MSCA, since that repo is being deprecated.

Sounds good @clpetersonucf I can probably help out, so just ping me when it’s ready

clpetersonucf:
In the Django rewrite we've transitioned to LTI 1.3, so this PR is now fundamentally incompatible. I'd like to revisit Sakai compatibility in the future, however.

PR #1433, opened 2023-01-17
Improved checks to $same_site config value

clpetersonucf:
By default, `$same_site` is null in the core Fuel config, but Fuel later attempts to cast the value to a string, converting `null` into an empty string. The checks in our overriden `cookie` class don't catch the empty string value, so `sameSame='None'` did not get applied correctly. The cookie class will now check for an empty string value and update `$same_site` to `'None'` manually.

### DIFF
--- a/fuel/app/classes/materia/fuel/core/cookie.php
+++ b/fuel/app/classes/materia/fuel/core/cookie.php
@@ -42,7 +42,7 @@ class Cookie extends Fuel\Core\Cookie
 		is_null($http_only) and $http_only = static::$config['http_only'];
 		//static::$config is protected - can't get it in an extended class, hack workaround here
 		// FuelPHP casts $same_site as a string, which converts null into an empty string. Just using is_null is insufficient.
-		if ( is_null($same_site) || ! $same_site) $same_site = 'None';
+		if ( ! $same_site) $same_site = 'None';
 
 		// add the current time so we have an offset
 		$expiration = $expiration > 0 ? $expiration + time() : 0;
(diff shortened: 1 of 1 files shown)


cayb0rg (approved):

fuel/app/classes/materia/fuel/core/cookie.php
```diff
@@ -41,7 +41,8 @@ public static function set($name, $value, $expiration = null, $path = null, $dom
 		// is_null($secure) and $secure = static::$config['secure'];
 		is_null($http_only) and $http_only = static::$config['http_only'];
 		//static::$config is protected - can't get it in an extended class, hack workaround here
-		is_null($same_site) and $same_site = 'None';
+		// FuelPHP casts $same_site as a string, which converts null into an empty string. Just using is_null is insufficient.
+		if ( is_null($same_site) || ! $same_site) $same_site = 'None';
```
FrenjaminBanklin, line 45: In hindsight, `is_null($same_site)` should also be covered with `! $same_site` so we may not even need both of them.

FrenjaminBanklin (approved):
One nit to pick aside, this looks all right to me.

Might also suggest going through the other null check + assignment lines in this function and switching them to explicit `if` checks, but if we're planning on moving away from PHP in the near future anyway it's probably fine as-is.

fuel/app/classes/materia/fuel/core/cookie.php
clpetersonucf: Removed `is_null` due to its redundancy.

PR #1434, opened 2023-01-17
Updated comparison used in usort function to return an integer value

clpetersonucf:
`usort` will error out in PHP 8.1 when the callback function returns a boolean instead of an integer. Updated the callback function used in the API's `score_summary_get` method to return an integer value instead.


FrenjaminBanklin (approved):
Confirmed this'll arrange semesters in reverse order. Does the job.

PR #1435, opened 2023-01-17
Added sequencer and syntax sorter to default widget list

clpetersonucf:
It's about time


FrenjaminBanklin (approved):
Sequencer and Syntax Sorter install properly on first run.

Seems good to me.

cayb0rg (approved):

PR #1436, opened 2023-01-17
Dev/9.0.3

clpetersonucf:
Contains several high priority resolutions to issues that have arisen due to the PHP 8.1 update.

- Fix to `SameSite=None` not being correctly applied to FuelPHP cookies.
- Fix to deprecated implementation of `usort` function in the `score_summary_get` API call.
- Added Sequencer and Syntax Sorter to the list of default widgets (finally).


FrenjaminBanklin (approved):
Nothing suspect in here, and we've verified all of these additions independently.

PR #1442, opened 2023-01-27
React media importer revisited

clpetersonucf:
This is a replacement of #1391, with unnecessary formatting changes stripped out and additional enhancements/fixes layered on top. It also includes several small bug fixes for the widget creator.

**Note: This PR includes a migration!**


fuel/app/classes/controller/api/asset.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * Materia
+ * License outlined in licenses folder
+ */
+
+use \Materia\Perm;
+use \Materia\Perm_Manager;
+
```
cayb0rg, line 9: Getting the error "Class 'Msg' not found" when user is not logged in. We should import \Materia\Msg here.

src/components/media-importer.jsx
```diff
@@ -0,0 +1,392 @@
...
+	const [assetList, setAssetList] = useState({})
+	const [showDeletedAssets, setShowDeletedAssets] = useState(false)
+	const [filterSearch, setFilterSearch] = useState('') // Search bar filter
+
+	const { data: listOfAssets } = useQuery({
+		queryKey: ['media-assets', selectedAsset],
+		queryFn: () => apiGetAssets(),
+		staleTime: Infinity,
+		onSettled: (data) => {
+			if (!data) console.warn('Error in asset retrieval')
```
cayb0rg, line 68: No login errors (401) are currently not being handled when trying to get, post, or delete an asset (could maybe use the Alert component here if data.title == “Invalid Login”?). apiAuthorVerify() in the creator does eventually handle this by showing the user as having been logged out due to inactivity.

fuel/app/classes/controller/api/asset.php
```diff
@@ -0,0 +1,58 @@
...
+			\DB::update('asset')
+				->value('deleted_at', time())
+				->value('is_deleted', '1')
+				->where('id', $asset_id)
+				->execute();
+		}
+		catch (\Exception $th)
+		{
+			trace('Error: In the deletion process');
+			trace($th);
```
cayb0rg, line 32: Not sure if we should return an error message here.

src/components/media-importer.jsx
```diff
@@ -0,0 +1,392 @@
...
+	// Asset list, sorting, search filter, or show delete flag is updated
+	// Processes the list sequentially based on the state of each
+	useEffect(() => {
+		if (!assetList || !assetList.length) return
+		
+		// first pass: filter out deleted assets, if we're not displaying them
+		let listStageOne = assetList.filter((asset) => (!showDeletedAssets && !asset.is_deleted) || showDeletedAssets )
+
+		// second pass: filter assets based on search string, if present
+		let listStageTwo = filterSearch.length ? listStageOne.filter((asset) => asset.name.toLowerCase().match(filterSearch)) : listStageOne
```
cayb0rg, line 115: filterSearch is not being converted to lower case in addition to asset.name (searching in uppercase will not return any results)

cayb0rg:
The key functionalities work nicely for me, styling changes look sweet, and tab accessibility is 👌. 

Wondering if we should have some way of telling the user (in AssetCard or as a popup) whether the selected asset type matches the REQUESTED_FILE_TYPES parameter. As of right now, it allows the user to select a previously uploaded asset of the incorrect file type to the creator.

src/components/media-importer.jsx
clpetersonucf: Added an additional check to see if the `data` payload contains an error, but otherwise deferred to the creator to actually display an error. Replaced the "Logged out due to inactivity" error with a more generic "You are no longer logged in" message to handle multiple use cases.

fuel/app/classes/controller/api/asset.php
clpetersonucf: You're right, the request will now return an error message consistent with other API requests.

src/components/media-importer.jsx
clpetersonucf: Good catch, resolved :+1:

FrenjaminBanklin (approved):
Still a lot of seemingly extraneous requests for thumbnails etc. on the server side after deleting anything from the media list, but everything seems to be working properly. No more console errors, and it looks like the duping I was noticing was probably just because there are multiple copies of certain media with the same name.

Looks like it's good enough to run for real.

clpetersonucf:
Merging, with the consideration that there may be some optimizations to be made later.

PR #1447, opened 2023-02-17
Resolves ZipArchive deprecated references

### LINKED ISSUE #1444: Deprecated ZipArchive reference in the Session_PlayDataExporter class
The `Session_PlayDataExporter` utilizes the `ZipArchive` utility to create zip files of CSVs of various score data to be exported upon request by users. As of PHP 8.1, ZipArchive is emitting the following fatal error:

```
ZipArchive::open(): Using empty file as ZipArchive is deprecated
```

This is apparently related to our use of `tempnam()`, creating an empty file as the source of the zip archive. Both the built-in play data exporter class will need to be updated, along with any widgets that have their own playdata exporters that includes zip file creation.

clpetersonucf:
Fixes #1444 by adding `ZipArchive::OVERWRITE` to `$zip->open($temname);` references in the `Session_PlayDataExporter` class. Note that widgets that have custom export options will also need this change implemented in their own play data exporters.

### DIFF
--- a/fuel/app/classes/materia/session/playdataexporter.php
+++ b/fuel/app/classes/materia/session/playdataexporter.php
@@ -334,7 +334,7 @@ class Session_PlayDataExporter
 		$tempname = tempnam('/tmp', 'materia_raw_log_csv');
 
 		$zip = new \ZipArchive();
-		$zip->open($tempname);
+		$zip->open($tempname, \ZipArchive::OVERWRITE);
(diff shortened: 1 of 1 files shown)


clpetersonucf:
As far as I can tell, `tempnam()` actually does the legwork of generating a unique filename on disk. The second argument is defined as the `prefix`, to which a random value is appended.

Marking as ready for review.

PR #1417, opened 2022-11-15
Multiple issues

emanuelGitCodes:
- Merge the ucf support react branch 
- Migration for deleted assets
- Remove packages from `package.json`
- Runs on node ver 14.20.0
- Score page, fix display order of attempts
- Rewrote `bar-graph.jsx` to remove jQuery
  - Applied to  Compare With Class
  - Applied to Student Activity in my-widget-page

Still working on writing test cases.

PR #1448, opened 2023-02-17
Implements Emanuel's changes to score screen and distribution graphs

clpetersonucf:
This implements changes Emanuel originally authored in #1417, independent of some of the other sweeping changes on that PR. It should fix the score distribution graph in the score summary and score screen.

**Note:** make sure you `rm -rf node_modules && yarn install` to implement changes to the `package.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,12 +22,7 @@
     "prettier:detectchanges": "git diff --exit-code ./src || (echo '!! Prettier created files that need to be manually added.'; exit 1;)"
   },
   "dependencies": {
-    "@vx/axis": "^0.0.199",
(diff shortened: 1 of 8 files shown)


clpetersonucf:
Two issues I'm noticing immediately:
- In the previous attempts drop-down, the order of the scores is correct (earliest at the bottom of the list) but the indexes are backwards. Ex: for 5 attempts, attempt 5 should be at the top (most recent) and attempt 1 at the bottom (earliest).
- For custom score screens, the header section is updated when selecting an earlier attempt, but the score screen does not update with the corresponding attempt's score details.

PR #1450, opened 2023-02-21
Removes redundant call to apiGetQuestionSet from scorescreen.

cayb0rg:
This issue was raised within [Guess The Phrase](https://github.com/ucfopen/guess-the-phrase-materia-widget/issues/30). I believe this may solve potential expiration issues, since the call to function question_set_get will sometimes cause the score screen to expire unnecessarily if the user is not authenticated. apiGetWidgetInstance has a built-in option to fetch the qset already, so I've made the score screen use that instead.


clpetersonucf (changes requested):
This looks good and does the job as expected - the only thing I'd like to see, and only because we have tests for the `widget_instances_get` API call already already, is to add additional tests for non-authenticated users requesting specific instance information, and to test the `$load_qset` flag.

clpetersonucf (approved):
This does what's needed - looking back at the big picture, allowing score screens to be viewed without being logged in AND supporting custom score screens with access to the qset makes for a tricky situation, but definitely a bigger one than what this PR was designed to solve. We might revisit it in the future, but this will serve for now.

Issue #1438, opened 2023-01-18
support_user role is missing some required permissions


clpetersonucf:
Support users need to have the ability to manage basic settings in the user and instance admin panels, namely:

- Editing a widget, making a copy, adding collaborators, adding extra attempts, deleting a widget
- Changing access settings like Guest Mode
- Changing availability settings

Currently, support_users cannot change some of these. The permissions requirements need to be reviewed.

clpetersonucf:
Resolved via #1456

PR #1456, opened 2023-03-07
Issue/1438 Grants support users additional permissions

cayb0rg:
Issue #1438 

After digging around, the lowest tier of access I could find that allows users to edit widget settings and grant access to other users was ownership permissions, or Perm::FULL. I could be wrong about this, so let me know what you think!

### NOTE: This PR includes a migration.

This PR also fixes an issue in the Collaborate dialog where users would appear more than once in the search results.

### DIFF
--- a/fuel/app/classes/materia/perm/manager.php
+++ b/fuel/app/classes/materia/perm/manager.php
@@ -665,9 +665,9 @@ class Perm_Manager
 
 		$all_perms = self::get_all_users_with_perms_to($object_id, $object_type);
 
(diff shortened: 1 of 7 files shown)


clpetersonucf (approved):
I'm going to approve and merge this, since it appears to check all the boxes. One behavior I'm noticing is an inability to view the score screen of a specific play ID for another user, which should probably be a permission available to support users. However, since that's slightly beyond the original scope of this PR, I'm fine making an issue for it to be tackled independently.

PR #1460, opened 2023-03-15
Issue/1459 remove asset hash cache busting

### LINKED ISSUE #1459: [React branch] Remove & (eventually) replace asset cache busting
There's an old implementation of asset cache busting via `process_assets.js` and `alwaysload.php` in the controllers. They generate a file at `fuel/app/config/asset_hash.json` that includes hashes for every asset served. As far as we can tell, these hashes are not put into practice: static assets served to the page do not include the hashes located in `asset_hash.json`. Their inclusion messes with the docker image build and release process, so they should probably be removed and eventually replaced with a different implementation of cache busting.

Note that for the purposes of the react branch, at least initially, the replacement solution is not expected.

clpetersonucf:
Resolves #1459 

Removes references to the dynamically compiled `asset_hash.json`, the hash processing code in `alwaysload.php`, and the `process_assets.js` file since it's no longer needed with the react branch.


FrenjaminBanklin (approved):
This runs locally.

Can't really 100% verify it myself, but assuming it also runs properly in CI/CD processes and builds a usable package for QA/Production use, willing to call it good.

PR #1466, opened 2023-03-29
Perms definitions update

clpetersonucf:
There's a lot of inconsistencies with group level permissions in the master branch:

- Group level permissions are defined but not really used except in one or two instances
- Group level permission definitions do not match current roles
- Group level permissions (espcially for superusers) are applied nonsensically

To fix this, this PR:

- Redefines the group level definitions to match current user roles
- Applies group level definitions to `support_user` where they did not exist on the react branch previously
- Updates the superuser permission level to the correct value

### DIFF
--- a/fuel/app/classes/materia/perm.php
+++ b/fuel/app/classes/materia/perm.php
@@ -62,10 +62,10 @@ abstract class Perm
 	const GIVE_SHARE   = 75;
 
 	// group rights only
-	/** @const Has rights to access manger interface */
-	const AUTHORACCESS  = 80;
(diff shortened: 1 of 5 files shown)


FrenjaminBanklin (approved):
These migrations run. I don't see any real issues.

Any effort put towards reducing the query count is probably not worth it given how simple these operations are.

The only thing I can think of maybe doing otherwise is adding the `Perm::BASICAUTHOR` to the `basic_author` role, but since that was never done before and nothing seems to use it, there's no real point.

Issue #1465, opened 2023-03-23
[React Branch] Settings dialog improvements


clpetersonucf:
- [Bug] The attempts slider does not allow any value to be set other than 'Unlimited' (the slider resets to the unlimited position after being dragged to another value)
- Saving the settings dialog re-fetches the instance list - this shouldn't be required, can it be improved?

clpetersonucf:
Resolved, with caveats that the slider behavior can still be improved. I'll leave this open for now.

PR #1468, opened 2023-04-05
React fixes and polish, part four

clpetersonucf:
- Resolves bug with collaborator count in My Widgets
- Styling improvements and fixes
- Adds a contextual footer to the widget player, with a link to the Player Guide (if available) and Materia
- Fixes contextual return URL in widget creator
- Fixes inoperable cancel dialog in save history importer
- Fixes styling errrors in collaborator dialog with long names
- Rewords widget sharing text for additional clarity
- Fixes attempt slider in the selected widget settings dialog, as mentioned in #1465 
- Fixes error related to deleting a widget instance
- Significant changes to my-widgets-page component and how it handles loading of instances. The selected widget instance should load as soon as it's presented in the paginated instance list. The throbber has moved to the top of the instance list to accommodate the change. Includes tweaks to other elements of loading behavior and adds a listener to the widget hash in the URL.
- Updated the `useUpdateWidget` mutation that handles widget settings updates to no longer invalidate the widget instance list. This prevents the list from manually refreshing any time a widget setting is changed.
- Adjustments to collaboration dialog to try and resolve type mismatch of user ID values in perms (thanks PHP 8.1!) and made sure the my-widgets page state correctly updated if you restrict your own access, as referenced in #1470 
- The `Cancel` button in the save history dialog had a tendency of wandering off when the dialog had y-overflow, it should now correctly remain where it should be


cayb0rg (approved):
Can't see anything wrong with it 👍 . Really like the loading icon in My Widgets and new footer in player!

FrenjaminBanklin (approved):
The few remaining instances of inline styles could maybe be replaced with classes, but otherwise it looks like this is all working well.

Issue #1469, opened 2023-04-11
[React Branch] The header is not currently displaying notifications


clpetersonucf:
As the title implies, the UI for notifications is not present in the header.

clpetersonucf:
It's there now.

Issue #1462, opened 2023-03-17
[React Branch] QA Feedback Master Issue


clpetersonucf:
Figured it'd be easier to create a single issue for people to submit feedback based on testing Materia v10 on QA. I can stub out issues based on items people submit.

clpetersonucf:
1. Widget request pagination is fairly slow in My Widgets, and makes it pretty clear 10 per request is far too few. That number could safely be bumped up considerably.
2. The `-cores` aren't loading properly, which is expected. NGINX dev was providing a redirect, but the easier thing is to probably scrap putting compiled assets in `public/dist` and just emit them straight into the `public` directory, to better match where those assets were located before.
3. We still have to fix the issue where collaborators are being displayed in duplicate due to a variable type change from the PHP 8.1 update.
4. Views that should be sourced from `Materia-Theme-UCF` are not rendering yet

clpetersonucf:
1. The creator Save History dialog's cancel button is operable
2. The widget icon alignment in the detail page is skewed
<img width="473" alt="Screenshot 2023-03-28 at 12 10 42 PM" src="https://user-images.githubusercontent.com/1268547/228302507-fe22fe6c-f2e4-423c-8beb-6b59f3ed175a.png">

clpetersonucf:
1. Embed code isn't generating correct width/height values for a given widget instance
2. Embedded widgets may not be initializing properly - this should be tested with live instances, and not just local dev instances embedded in a local html file.

clpetersonucf:
New issues discovered in the apha4 build:
- If a widget hash is present in the URL, the widget will not be selected when the instance list is fully loaded
- Copying or deleting a widget reloads the widget list, but the filter to sort them by creation date does not appear to be operating on the list
- The 'or create a new widget' button appears to be below the viewport of the LTI picker window in Canvas:
<img width="738" alt="Screenshot 2023-04-20 at 1 51 14 PM" src="https://user-images.githubusercontent.com/1268547/233447935-b89f27bd-4b1d-417d-86e0-d53bfa6a2f15.png">

cayb0rg:
* Searching by creation date/time doesn’t work in the admin/support pages
* (Missing feature) Cannot see the owner of a widget in admin/support pages (could have been omitted for privacy?)
* Help/Support contact details missing
* Widget navigation in My Widgets disappears on smaller screen widths (<980px):

[...]

PR #1473, opened 2023-04-25
Issue-1469/notifications

cayb0rg:
Starts the notification UI for #1469


clpetersonucf (changes requested):
It's mostly functional, a few quick comments:
Bugs:
- Dismissing a notification message will clear the message inbox but the envelope icon persists. Ideally the icon should be dismissed completely when the message queue is empty. Right now, the envelope icon continues to persist, even after logging out and logging back in.
- Actions performed through notifications are non functional. There's only one right now: granting access to another user if they request access via the LTI pre-embed placeholder. An easy way to test this:
1. Visit `/lti/test/provider` and select `LTI Picker Launch` -> `As Instructor` 
2. Select a widget to get the pre-embed confirmation.
3. Select `LTI Picker Launch` -> `As New Instructor`
4. Request access.
5. Launch via `As Instructor` again, then select the option to create a new widget in Materia.

You should have the notification requesting access. For me, granting access yields a console error. On production, it loads the widget in My Widgets, opens the collaborator dialog, and pre-populates requesting user as a collaborator. This may require some additional engineering to those components, so let me know if you want a hand with getting it working. 

Design feedback:
- Font scaling in the notification shade is a little all over the place. I would reduce all font sizes a bit, including the action button: Screenshot 2023-05-09 at 4 05 22 PM
- The close button is a PNG file that hasn't scaled well on high resolution displays. I would like to replace it with an SVG or font glyph that isn't rasterized.

Overall, good work so far :+1:

cayb0rg:
Thanks for the feedback!

New changes:
* “Grant access” now redirects user to my widgets and adds the requesting user in collaborator
* Notification shows error message on server failure
* Adds sent timestamp to notifications
* Fixes request access success message showing for all owners in LTI view
* Fixes widget link href in notifications
* Sender avatar is no longer clickable
* Changes delete button to an SVG
* Clicking outside notifications closes notifications
* Notification symbol no longer persists after notifications have all been dismissed

cayb0rg:
It does make more sense to perform the update after the player hits save. I've changed it to update the query data instead, but this may need further testing.

clpetersonucf (changes requested):
LGTM. I really like the new flow for granting access, great job!

The only remaining requested change: clear out the remaining console logs, I would also update the console.log associated with the `onError` in the `useSetUserInstancePerms` hook: 
```
onError: () => console.log('failed to update user perms')
```

with a `console.warn` or `console.error`

clpetersonucf (changes requested):
I was ready to merge this in but did a bit of final testing after resolving the merge conflict and found a few additional issues:

- The collab dialog is duplicating the last user in the list when receiving the data from `user_get` when in the context of the instance admin panel. The collab dialog in My Widgets doesn't appear to have this issue.
- Adding someone as a collaborator from My Widgets causes the page to error out as soon as a user is selected from the search results drop-down.

I'm not sure if these behaviors were present prior to resolving the merge conflict - resolving it had to do with renaming the state variable `updatedOtherUserPerms` to `updatedAllUserPerms` in the `my-widgets-collaborate-dialog` component.

clpetersonucf:
A follow-up to my last comment: the issues relating to duplicate collaborators is related to loose management of variable types, particularly with user perms.

In the `my-widgets-page` component, we're explicitly setting the user ID key of the other user perms' object to an int:
https://github.com/ucfopen/Materia/blob/0d139fafcecf8931b59f7d0da2c42ea118088004/src/components/my-widgets-page.jsx#L128

However, in the `support-selected-instance` component, that same value is not being explicitly set. It ends up being cast as a string, which is something that (iirc) changed with the PHP 8.1 update. That causes comparisons downstream to fail, like this one, which is why collaborators end up duplicated in one version of the collaborator dialog but not another:
https://github.com/cayb0rg/Materia/blob/69f311cd0e1ea44381f9575bc1cbd6d4252f5c6a/src/components/my-widgets-collaborate-dialog.jsx#L250

I would wrap the user perm key in a `parseInt()` in the support-selected-instance component as well, because IMO, ints are preferable to strings for these numeric IDs. That also means I'd probably remove this line:
https://github.com/cayb0rg/Materia/blob/69f311cd0e1ea44381f9575bc1cbd6d4252f5c6a/src/components/my-widgets-collaborate-dialog.jsx#LL83C3-L83C3
Since it's trying to perform the opposite. Unfortunately, the looseness of type definitions in javascript is biting us here a little bit 😬

cayb0rg:
New changes:

* Deletes notifications optimistically (instead of calling notification_get every time a notification is deleted)
* Support users can be added as collaborators even if they are students
* Adds null check for newCollabUser
* Fixes an issue with added users not appearing in collaborator
* Disabled option to remove yourself as a collaborator if you have view access (quick fix to [this issue](https://github.com/ucfopen/Materia/issues/1462#issuecomment-1570780265))
* Adds parseInt() to support selected instance to fix duplicated users

PR #1482, opened 2023-06-06
Fixes to notifications and collaborator

cayb0rg:
* In the case where a user sends an access request twice and the recipient grants them Full Access in one notification, and then grants them View Scores access in the next notification (or vice versa) without leaving the page, the collaborator dialog will still show the user as having Full Access. Just needed to reverse the merge priority when creating a JavaScript map from otherUserPerms and state.updatedAllUserPerms.
* Fixes notification count on envelope icon not decrementing after granting a user access.
* Removes leftover code from from testing the setQueryData
* Users can now remove themselves as collaborators regardless of their level of access.

### DIFF
--- a/fuel/app/classes/model/user.php
+++ b/fuel/app/classes/model/user.php
@@ -94,7 +94,7 @@ class Model_User extends Orm\Model
 		$user_table = \Model_User::table();
 		$matches = \DB::select()
 			->from($user_table)
-			// Do not return super users or the current user
+			// Do not return super users or the current user // << why?
 			->where($user_table.'.id', 'NOT', \DB::expr('IN('.\DB::select($user_table.'.id')
 				->from($user_table)
 				->join('perm_role_to_user', 'LEFT')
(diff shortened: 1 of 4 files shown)


fuel/app/classes/model/user.php
```diff
@@ -94,7 +94,7 @@ static public function find_by_name_search($name)
 		$user_table = \Model_User::table();
 		$matches = \DB::select()
 			->from($user_table)
-			// Do not return super users or the current user
+			// Do not return super users or the current user // << why?
```
clpetersonucf, line 97: This is because:
- superusers aren't supposed to be visible to normal users (the superuser account isn't associated with a UCF identity)
- Previously, the only time this search was run was in the context of sharing a widget with someone, so there was no need to search for yourself. With the addition of instance admin, it makes sense to include the current user in the search results (so you can add yourself to another widget, for example), so this could probably be changed. That's not really within the scope of this PR though.

clpetersonucf (approved):

PR #1483, opened 2023-06-12
[React branch] Improved pagination behavior in My Widgets

clpetersonucf:
- Replaced the previous query for the paginated instances list with an infinite query
- Wrapped instances query in a custom hook that's separated from the my-widgets-page component
- Modifications to the `widget_paginate_instances_get` API call and its associated underpinnings in the instance manager
- Reworked the `useCopyWidget` mutation to correctly treat the query cache as immutable, no longer invalidates the `widgets` query cache
- `useDeleteWidget` mutation updated for new pagination structure
- Changes to widget copy API endpoint to support in-place updates of the widget list after copying, preventing the need to to re-request the instance list from the server
- Cleanup of the `my-widgets-page` component associated with the above
- Updated the lti `select-item` component to make use of the new `useInstanceList` hook

### DIFF
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -60,13 +60,13 @@ class Api_V1
  * Takes a page number, and returns objects containing the total_num_pages and
  * widget instances that are visible to the user.
  *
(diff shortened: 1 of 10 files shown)


cayb0rg (approved):
LGTM.

PR #1487, opened 2023-07-27
Catalog accessibility indicator updates

clpetersonucf:
- Splits widget catalog filter options into Features and Accessibility. The options can be displayed separately or in combination, and the filters are applied in aggregate.
- Update accessibility UI elements in widget detail pages.
- Adds an additional `accessibility_description` `meta_data` field to provide additional context about a widget's accessibility status. The indicators are no longer displayed if a widget does not provide accessibility information; the description will simply note that accessibility data is not available.

### DIFF
--- a/src/components/accessibility-indicator.jsx
+++ b/src/components/accessibility-indicator.jsx
@@ -15,12 +15,20 @@ const AccessibilityIndicator = ({widget = {}}) => {
 		}
 	}
 
+	let descriptionRender = ''
+	if (widget.accessibility.keyboard.toLowerCase() != 'unavailable' || widget.accessibility.screen_reader.toLowerCase() != 'unavailable') {
+		descriptionRender = widget.accessibility.description
+	}
+	else {
+		descriptionRender = 'No accessibility information is provided for this widget.'
+	}
+
 	return (
 		<div className='feature-list accessibility-options'>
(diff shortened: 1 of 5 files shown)


cayb0rg:
Widget detail page looks good, accessibility filters work great. Just needs some mobile optimization: the pseudo-elements for Features and Accessibility get overlapped on screen sizes < 960px, and accessibility filter completely disappears on screen sizes < 720px.

cayb0rg:
If possible, we should keep the labels "Filter by feature" and "Filter by accessibility" after a filter is checked on smaller screen sizes.

image


Other than that, this looks good to go.

FrenjaminBanklin (approved):
This looks good and functions well, I can only think of two pretty minor things that could maybe be improved:

- The ARIA labels for the toggles aren't very descriptive, something like 'display filter options by feature' when toggling on or 'hide filter options by feature' when toggling off etc. would probably do. Or potentially a more verbose instructive label since there isn't much in the way of explaining how to use the filters.
  - Likewise the filter options themselves could probably use a more descriptive label so it's more immediately clear what they do.
- The filter options are all reachable in the tab order whether they've been toggled or not.

I wouldn't hold up the initial release of these features for either of these unless they're super quick fixes, though.

PR #1488, opened 2023-08-14
Fixes refresh in LTI picker and widget player iframe scrolling

cayb0rg:
Refetches query in ``useInstanceList`` on manual refresh in the LTI picker

Adds footer height to the embedded iframe's height and updates footer styling so the scrollbar is removed in embedded widgets


clpetersonucf (approved):
You beat me to making these changes :+1: Nice work

PR #1490, opened 2023-08-16
Fix widget footer margins

cayb0rg:
Prevents widget footer from centering itself in the page.

### DIFF
--- a/src/components/widget-player.scss
+++ b/src/components/widget-player.scss
@@ -92,8 +92,6 @@ section.widget {
 
 	section.player-footer {
 		height: 38px;
-
-		margin: 0 auto;
 		padding: 4px 10px;
 
 		border-top: solid 2px $color-purple;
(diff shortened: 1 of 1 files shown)


clpetersonucf:
This works, but I'd actually prefer to add `margin: 0 auto;` to the `section.widget .center` selector in the same file. Currently, the fix causes the player to left-align to the available space, while the alternative would center-align the player and footer together (but we should still test the player in various embed environments to make sure it displays properly).

cayb0rg:
True. I considered that as well and thought being left-aligned looked more congruent with WebCourses, especially on wide screen sizes. If the author wanted to center one embedded in a page, they could use WebCourses' text centering option (or add ``style = "text-align: center;"`` if it's outside WebCourses). However, I'm totally open to making it centered by default since this centering option is not available for LTI embedded widgets, and there are ways of making it left-aligned again in pages if they wanted it to be.

clpetersonucf (approved):

PR #1495, opened 2023-09-21
React branch media importer updates

clpetersonucf:
- Fixed previously imported media assets not being filtered in the importer based on allowed file types requested by the widget
- Added support for .wav and .m4a audio formats
- Rebuilt file upload logic to be faster and more performant by removing dependency on `FileReader` conversion to a Data URI followed by a Blob.
- Added language indicating allowed media type(s)
- Added additional error states
- Added upload progress indicator

### DIFF
--- a/fuel/app/classes/materia/widget/asset.php
+++ b/fuel/app/classes/materia/widget/asset.php
@@ -20,6 +20,7 @@ class Widget_Asset
 		'image/gif'   => 'gif',
 		'image/jpeg'  => 'jpg',
 		'audio/mp4'   => 'm4a',
+		'audio/x-m4a' => 'm4a',
 		'audio/mpeg'  => 'mp3',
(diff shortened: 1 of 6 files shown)


cayb0rg (approved):
This looks good to me! 👍🏻 Only issue I encountered was trying to upload .m4a files (Firefox and Chrome both recognized them as audio/x-m4a), but that could just be an issue on my end.

clpetersonucf:
@cayb0rg I didn't see any mention of the `audio/x-m4a` MIME type in my (admittedly limited) research. Try again?

cayb0rg:
Getting the same issue. It's not a file type I'm familiar with, but I think audio/x-m4a is just an alias for audio/m4a. If you're able to upload .m4a files with no issue, I'd say let it slide.

image

PR #1496, opened 2023-09-25
Add expected `img` property to widget instance objects used by the LTI widget picker.

FrenjaminBanklin:
Removes `iconUrl` use from widget picker, using it instead in the `useInstanceList` hook to add the `img` property to all widget instance objects for continued compatibility with any LTIs expecting that property such as Obojobo.

Fixes the assignment of the `outcome_ext` key in the LTI launch variable parsing code where it was erroneously defaulting to `null` instead of an empty string in the event that `ext_outcome_data_values_accepted` is not provided by incoming LTI params.

### DIFF
--- a/src/components/hooks/useInstanceList.jsx
+++ b/src/components/hooks/useInstanceList.jsx
@@ -27,7 +27,7 @@ export default function useInstanceList() {
 						//  compatibility with any downstream LTIs using the widget picker
 						return {
 							...instance,
-							img: iconUrl(BASE_URL + 'widget/', instance.widget.dir, 60)
+							img: iconUrl(BASE_URL + 'widget/', instance.widget.dir, 275)
 						}
 					}))
 				)
--- a/src/components/hooks/useInstanceList.jsx
+++ b/src/components/hooks/useInstanceList.jsx
@@ -1,6 +1,7 @@
 import { useState, useEffect, useMemo } from 'react'
 import { useInfiniteQuery } from 'react-query'
 import { apiGetWidgetInstances } from '../../util/api'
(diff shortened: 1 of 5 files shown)


src/components/hooks/useInstanceList.jsx
```diff
@@ -19,7 +20,18 @@ export default function useInstanceList() {
...
 			let dataMap = []
-			return [...dataMap.concat(...list.pages.map((page) => page.pagination))].sort(_compareWidgets)
+			return [
+				...dataMap.concat(
+					...list.pages.map(page => page.pagination.map(instance => {
+						// adding an 'img' property to widget instance objects for continued
+						//  compatibility with any downstream LTIs using the widget picker
+						return {
+							...instance,
+							img: iconUrl(BASE_URL + 'widget/', instance.widget.dir, 60)
```
clpetersonucf, line 30: This works fine, my only suggestion might be to use 275 instead of 60 for the icon size. We're also making use of `iconUrl` in the `my-widgets-instance-card` component, and by using 275 we can effectively replace both `iconUrl` references with a reference to the `img` property of the instance. That line is here: https://github.com/ucfopen/Materia/blob/issue/support-dashboard-in-react/src/components/my-widgets-instance-card.jsx#L30

I also noticed this does indeed fix the icon not being displayed during the progress bar view of the `select-item` component, but it's missing some styles to size and position it correctly. I'm happy to fix that after the merge if you prefer.

clpetersonucf (changes requested):
One small recommendation, otherwise it looks good and does the job.

clpetersonucf (approved):
Nice. LGTM

PR #1497, opened 2023-09-26
[React branch] Widget player component refactor

clpetersonucf:
- Heavy refactor of the widget player component, to address potential race conditions with logging and streamline various bits of the logic associated with handling, enqueuing, and submitting play logs.
- Fixed improper type matching of the `is_valid` property of `Session_Play`
- Implemented missing `session_play_verify` heartbeat associated with active widget plays

### DIFF
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -1165,7 +1165,7 @@ class Api_V1
 		{
 			if ($play->get_by_id($play_id))
 			{
-				if ($play->is_valid == '1')
+				if (intval($play->is_valid) == 1)
 				{
 					$play->update_elapsed(); // update the elapsed time
 					return $play;
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -1165,7 +1165,7 @@ class Api_V1
 		{
 			if ($play->get_by_id($play_id))
 			{
-				if ($play->is_valid == 1)
+				if ($play->is_valid == '1')
 				{
 					$play->update_elapsed(); // update the elapsed time
 					return $play;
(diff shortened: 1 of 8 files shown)


src/components/widget-player.jsx
```diff
@@ -249,26 +249,20 @@ const WidgetPlayer = ({instanceId, playId, minHeight='', minWidth='',showFooter=
 				if (isPreview) {
 					args.push(inst.id)
 				}
-				const newQueue = [{ request: args }]
+				var newQueue = [{ request: args }]
```
FrenjaminBanklin, line 252: Does this assignment have to happen, or could we save the line (and the use of `var`) by just doing this as we call `_pushPendingLogs`?

Not worth holding anything up over, but it would be cleaner.

src/components/widget-player.jsx
```diff
@@ -249,26 +249,20 @@ const WidgetPlayer = ({instanceId, playId, minHeight='', minWidth='',showFooter=
...
-				saveStorage.mutate({
-					play_id: playId,
-					logs: pendingLogs.storage,
-					successFunc: () => {
-						dispatchPendingLogs({type: 'clearStorage'})
-						setQueueProcessing(false)
-					}
-				})
+				const args = [playId, pendingLogs.storage]
+				var newQueue = [{ request: args }]
```
FrenjaminBanklin, line 259: Same as the above comment.

cayb0rg (approved):

fuel/app/classes/materia/api/v1.php
```diff
@@ -1165,7 +1165,7 @@ static private function _validate_play_id($play_id)
 		{
 			if ($play->get_by_id($play_id))
 			{
-				if ($play->is_valid == 1)
+				if ($play->is_valid == '1')
```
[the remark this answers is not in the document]
clpetersonucf: I was thinking the string check was probably too likely to cause problems. Your solution is better. Adopted.

src/components/widget-player.jsx
clpetersonucf: I agree, more of a holdover from how the component was originally written. Updated.

PR #1498, opened 2023-09-27
React Branch fixes and polish part seven

clpetersonucf:
Addressing final (!!!!?) issues before the update goes live:

- Fix to score overview not properly updating, tweaked prev attempt display to reduce component rerenders
- Updates to styles in widget creator: Previewing Prior Save mode for the action bar, and the publish/preview pop-up
- README updates and cleanup
- Replaced favicon, added multiple favicon sizes
- Optimized several images to reduce filesize
- Reworked LTI navigation landing page content to be better presented and more relevant
- Added Student Activity variant for widgets that don't collect scores
- Tweaks to Export Options dialog in My Widgets


FrenjaminBanklin (approved):
Looks like everything's working well.

cayb0rg (changes requested):
Looks good! Just ran into one issue on score screen: selecting an attempt or changing the attempt hash will render it on the page for a second before it flops back to the original attempt hash.

FrenjaminBanklin (approved):
Probably should've tried to get different scores on different attempts so the score screen actually changes visibly.

It looks like choosing a different attempt is correctly updating the score screen now.

PR #1500, opened 2023-09-29
[React Branch] Fixes and polish: final edition

clpetersonucf:
- Corrects strict equality check causing window `IS_EMBEDDED` to not parse correctly on the scores page
- Resolves docker compose v2 backwards incompatibility when running tests, including the "Input device is not a TTY" error and the test suite being incorrectly run against the primary database
- Updates `Materia-Theme-UCF` to 2.0.0 in composer.json, updated composer lockfile
- `NoScoreContent` text updated for draft widgets to indicate draft widgets must be published to collect scores
- Small tweaks to edit widget warning modal styling
- Fixes to style and overflow issues with the collab dialog's search results
- Adjusted access settings rendering to accommodate an edge case when a student has access to a non-student widget
- Removed server session caching for support and super user roles, pending a better implementation

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "eher/oauth": "1.0.7",
         "aws/aws-sdk-php": "3.67.17",
         "symfony/dotenv": "^5.1",
-        "ucfopen/materia-theme-ucf": "2.0.0-alpha1"
+        "ucfopen/materia-theme-ucf": "2.0.0"
     },
     "suggest": {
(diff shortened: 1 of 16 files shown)


FrenjaminBanklin (approved):
Preventing student users who used to be author users from changing the access controls on normal-access widget instances might be safer in the long run, but it seems like it happens infrequently enough that it can be its own future issue.

Everything's running well and looking good.

PR #1501, opened 2023-10-03
Quick misc fixes

cayb0rg:
- Adds simple null checks on `data` before accessing properties on `data` (If `data` is null, that is usually a sign something went wrong on the server; this is just a front-end band-aid)
- Fixes notification error message from appearing in every notification row
- Adds widget link to request access notification
- Fixes time conversion in unavailable widget messages (#1509)
- Adds styling to Instance Manager inside User Admin
- Encodes apostrophes in admin instance search query

### DIFF
--- a/src/components/my-widgets-page.jsx
+++ b/src/components/my-widgets-page.jsx
@@ -215,7 +215,7 @@ const MyWidgetsPage = () => {
 				widgetName: inst.widget.name,
 				dir: inst.widget.dir,
 				successFunc: (data) => {
-					if (!data || (data.type == 'error'))
+					if (data && (data.type == 'error'))
 					{
 						console.error(`Failed to copy widget with error: ${data.msg}`);
 						if (data.title == "Invalid Login")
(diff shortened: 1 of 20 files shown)


src/components/widget-creator.jsx
```diff
@@ -103,14 +103,14 @@ const WidgetCreator = ({instId, widgetId, minHeight='', minWidth=''}) => {
...
 		enabled: !!instance.id, // requires instance state object to be prepopulated
 		onSettled: (data) => {
-			if ( (data != null ? data.title : undefined) === 'Permission Denied' ||data.title === 'error') {
-					setCreatorState({...creatorState, invalid: true})
-					onInitFail('Permission Denied')
-				} else {
-					setCreatorState({...creatorState, invalid: false})
-					setInstance({ ...instance, qset: data })
-				}
+			if ( (data != null ? data.title : undefined) === 'Permission Denied' || (data && data.title === 'error')) {
```
clpetersonucf, line 106: This check won't catch `data == null`, but the API should never send a null object for this request. If an instance qset isn't found it sends a dummy one instead, so this should be sufficient.

src/components/my-widgets-page.jsx
```diff
@@ -215,7 +215,7 @@ const MyWidgetsPage = () => {
 				widgetName: inst.widget.name,
 				dir: inst.widget.dir,
 				successFunc: (data) => {
-					if (!data || (data.type == 'error'))
+					if (data && (data.type == 'error'))
```
clpetersonucf, line 218: The lack of a null check here is probably fine since this is coming from the `useDeleteWidget` mutation which handles errors.

clpetersonucf (approved):
I don't see any issues here, approved

clpetersonucf (approved):
Re-upping approval.

Issue #1502, opened 2023-10-05
Collaboration dialog student && support permissions overlap


clpetersonucf:
The collaboration dialog will seemingly allow you to grant a student access to a non-student widget (that's not in Guest Mode) if that student also has the support role. The dialog can be saved without error, and the API does not return a permissions error or it is not acted upon by the client. However, the permission update doesn't actually "stick": it's not updated in the database and refreshing the page will show that the user was not actually added.

Ideally we handle this situation more gracefully, likely by continuing to enforce the student role limitations despite the user being support.

PR #1515, opened 2023-10-05
Issue/1502

cayb0rg:
Triggers the Share Not Allowed warning dialog when attempting to add a student to a widget without guest access, even if that student has the support role. When attempting to add multiple users as collaborators, it will add all but students and return an error saying students cannot be added as collaborators.

Fix for #1502


clpetersonucf (approved):
LGTM. I'm holding off on merging since this will be targeted at the `dev/10.0.1` branch, which should be forthcoming.

PR #1516, opened 2023-10-05
Issue/1512-and-1504

### LINKED ISSUE #1504: Authors can potentially lose access to their own instances after becoming students
To reproduce, open the settings dialog and select "save". The permissions update kicks in and revokes the user's access to the instance immediately.

This could be a messy situation to resolve. Perhaps the instance should be auto-set to guest mode? Or perhaps the permissions update should ignore the awkward situation and maintain the users' current access if they were grandfathered in?

### LINKED ISSUE #1512: Improve access settings when a student views settings for a non-student widget
Related: #1504 

Right now we disable access settings in this context, but the Embedded Only option, if available, is not blurred and disabled.

cayb0rg:
Fixes #1512 
- Hides embedded only option for students and student made widgets and increases dialog's max-height

Fixes #1504 
- ~~Went with the option of setting all of the user's instances to guest mode, since a user whose author status has been removed having access to graded widgets feels icky~~
- Updates settings dialog view and edit permissions for grandfathered-in users

Additional
- Added error message for when you try to remove guest access from a student-made widget in instance manager
- Adds 'owner' tag to collaborator dialog

### DIFF
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -237,6 +237,10 @@ class Api_V1
 		// student made widgets are locked forever
 		if ($inst->is_student_made)
 		{
+			if ( $guest_access === false )
+			{
+				return new Msg('Student-made widgets must stay in guest access mode.', 'Student Made', 'error', false);
+			}
 			$attempts = -1;
 			$guest_access = true;
 		}
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -237,7 +237,7 @@ class Api_V1
 		// student made widgets are locked forever
 		if ($inst->is_student_made)
 		{
-			if ( $guest_access === false )
+			if ($guest_access === false)
 			{
 				return new Msg('Student-made widgets must stay in guest access mode.', 'Student Made', 'error', false);
 			}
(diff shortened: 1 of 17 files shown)


clpetersonucf (changes requested):
I appreciate the idea you had here - and I hate to put some cold water on it, but I think universally flipping all instances owned by the user to guest mode is too heavy-handed. The primary reason being that if a user once again receives the author role, we don't have a method to handle the inverse, and there may be widgets previously created that were set to guest mode intentionally that could be flipped to normal mode by accident.

Rather, I'm more inclined to retain all users' instance permissions, but make features and  permissions in My Widgets a little more locked down for instances that are **not** student-made when the user is a student.

Instances in this "grandfathered-in" mode could have the following changes, potentially:

- Attempt limits can ONLY be retained or modified to unlimited, and not any other value.
- Access can ONLY be set to Guest Mode, but the current access mode can be retained.
- Access to Student Activity is restricted (blocked)
- Access to Export Options is restricted (either blocked outright, or only Questions & Answers is available)

I think it's pretty safe to assume that for instances students already have access to but aren't student-made, the implication is that the owner was once not a student. These restrictions are pretty targeted without being overly penalizing or inconvenient. Let me know how you feel about it - I'm more than happy to go over these restrictions more in-depth or come up with an alternative.

clpetersonucf:
Deleting the `ucfopen/dev/10.0.1` force-closed this PR (sorry!). You'll have to remake it and target the new `dev/10.1.0` branch instead. I would add a reference to this PR so the prior history and comments can be tracked.

PR #1518, opened 2023-10-11
Issue/1508-pagination

### LINKED ISSUE #1508: User and Instance admin search results should be paginated
These can be chunky requests, especially with the low debounce time. Ideally the returned results are paginated and a progress indicator should be used.

cayb0rg:
Fixes #1508

* Adds pagination to the admin interface's instance search by piggybacking on the `useInstanceList` hook used in My Widgets Page
* Adds offset and limit to SQL query for instances in My Widgets Page
* Adds pagination to admin user search
  * This is also applied to the collaborator and extra attempts dialog
* Adds scrollbar to collaborator dialog user search results
* Set avatar sizes in user admin search to 50x50

### DIFF
--- a/.github/workflows/test_and_build.yml
+++ b/.github/workflows/test_and_build.yml
@@ -53,7 +53,7 @@ jobs:
           cd docker && ./run_build_github_release_package.sh ghcr.io/${{ github.repository_owner }}/materia:app-${{ steps.tag_name.outputs.GIT_TAG }}
 
       - name: Upload to Release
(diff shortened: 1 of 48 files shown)


clpetersonucf:
Same as #1516, deleting the ucfopen/dev/10.0.1 force-closed this PR (sorry!). You'll have to remake it and target the new dev/10.1.0 branch instead. I would add a reference to this PR so the prior history and comments can be tracked.

PR #1521, opened 2023-10-12
Dev/10.0.1

### LINKED ISSUE #1502: Collaboration dialog student && support permissions overlap

### LINKED ISSUE #1505: Enhance the help section of the main site
The help section of the site could use a few additions to address general concerns/issues, including:

For players:
- Why didn't my score return to Webcourses?
- I'm having an issue with the widget
- What can I do in Materia as a student?

For instructors/creators:
- How to share widgets with other collaborators
- Accessing widgets you don't own in Webcourses
- How do widgets save, is it automatic?
- What are my options for sharing a widget with students?

Potentially, the Materia direct link in the player footer can link to the expanded player help section.

### LINKED ISSUE #1509: Instance open/close time parsing can be invalid
For example, if I enter `1300` it should predict that means 1pm: right now it correctly gets set to `1:00`, but the PM toggle is not automatically updated. Additionally, the time displayed for the widget's play preview ("this instance is only available after XX:YY" or until XX:YY") doesn't always match the time value entered in the settings dialog.

### LINKED ISSUE #1523: Paginated score objects are not being merged correctly in individual student scores
In the `MyWidgetScoreSemesterIndividual` component, `state.logs` is an array where every index is the User ID of a student. `state.logs[id].scores` should be a flat array of all the scores for a given instance for a given student, but when multiple pages of results are returned, the `scores` array can instead become an object nested within an array, causing problems with the component rendering.

clpetersonucf:
`dev/10.0.1` is the primary dev branch associated with the forthcoming `v10.0.1` release.

- fixes #1502 by restricting student access to non-student widgets even if they have the support user role
- MWD package config now includes required `path.js`
- Adds simple null checks on data before accessing properties on data (If data is null, that is usually a sign something went wrong on the server; this is just a front-end band-aid)
- Fixes notification error message from appearing in every notification row
- Adds widget link to request access notification
- Fixes time conversion in unavailable widget messages (resolves #1509)
- Adds styling to Instance Manager inside User Admin
- Encodes apostrophes in admin instance search query
- Expands and enhances help section to include common issues for students & instructors (resolves #1505)
- Fixes issue with merging paginated play score data in the individual scores component (resolves #1523)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "eher/oauth": "1.0.7",
         "aws/aws-sdk-php": "3.67.17",
         "symfony/dotenv": "^5.1",
-        "ucfopen/materia-theme-ucf": "2.0.1"
+        "ucfopen/materia-theme-ucf": "2.0.2"
     },
     "suggest": {
         "ext-memcached": "*"
@@ -97,9 +97,9 @@
             "package": {
                 "name": "ucfopen/materia-theme-ucf",
                 "type": "fuel-package",
-                "version": "2.0.1",
+                "version": "2.0.2",
                 "dist": {
-                    "url": "https://github.com/ucfopen/Materia-Theme-UCF/archive/refs/tags/v2.0.1.zip",
+                    "url": "https://github.com/ucfopen/Materia-Theme-UCF/archive/refs/tags/v2.0.2.zip",
                     "type": "zip"
                 },
                 "source": {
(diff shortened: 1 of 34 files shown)


cayb0rg (approved):
Avatars in user admin results are missing width and height styling. Otherwise, I didn't find any related issues.

dgwn (approved):
Individual PRs were vetted and the PR was previously approved 👍

PR #1524, opened 2023-10-16
Fixes bug in merging scores from multiple query pages

### LINKED ISSUE #1523: Paginated score objects are not being merged correctly in individual student scores

clpetersonucf:
Resolves #1523


cayb0rg (approved):

FrenjaminBanklin (approved):
Solves the problem cleanly.

Looks good to me.

PR #1538, opened 2023-10-26
Audio recording in media importer

clpetersonucf:
Adds a built-in media recorder option to the media importer.

Note: this uses the [mediaRecorder](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder) interface which is relatively new. It has cross-browser support, but a few quirks: there may be browser compatibility issues related to the audio format(s) used for recording. `audio/webm; codecs=opus` is hard-coded as the MIME type due to general compatibility.

Additionally, the blobs generated by the `mediaRecorder` interface due not contain duration metadata. This is apparently [a bug in chromium](https://issues.chromium.org/issues/40482588) that has previously been tagged as `wontfix` due to it being "by design". Instead, I had to use a [fix-webm-duration](https://github.com/yusitnikov/fix-webm-duration) package that reads the blob after recording has stopped and returns a modified blob that includes the correct duration information.

Replacement for #1527 since that was accidentally closed when I deleted `dev/10.0.1` before changing the original PR's base branch.


clpetersonucf:

PR #1540, opened 2023-10-31
Issue/settings dialog

### LINKED ISSUE #1504: Authors can potentially lose access to their own instances after becoming students

### LINKED ISSUE #1512: Improve access settings when a student views settings for a non-student widget

cayb0rg:
Fixes #1512 
- Hides embedded only option for students and student made widgets and increases dialog's max-height

Fixes #1504 
- Updates settings dialog view and edit permissions for grandfathered-in users

Additional
- Added error message for when you try to remove guest access from a student-made widget in instance manager
- Adds 'owner' tag to collaborator dialog

Continuation of #1516


src/components/my-widgets-scores.jsx
```diff
@@ -61,7 +61,9 @@ const MyWidgetsScores = ({inst, beardMode}) => {
 	const handleShowOlderClick = () => setState({...state, isShowingAll: !state.isShowingAll})
 
 	let contentRender = <LoadingIcon />
-	if (isFetched) {
+	if (is_student) {
+		contentRender = <p>Students cannot view scores.</p>
```
clpetersonucf, line 65: Could we modify this so that students can only view scores if the widget is in guest mode? Historically, we've been fine with student users being able to view score results, just anonymized ones. An even better solution would be to potentially filter out score results except for "All Guests", in edge cases where a student is given access to a widget that previously had non-anonymized scores.

clpetersonucf (changes requested):
Great work. I left one request below based on displaying scores, but I'd also like to see some level of responsiveness added to the settings dialog. Currently, if a user is interacting with the dialog with a lower resolution display, the bottom of the dialog is cut off and prevents them from selecting the Save and Cancel options at the bottom.

cayb0rg:
I overlooked the fact that students should still be able to view guest scores, thank you! I added an `is_student` check to `play_logs_get` to exclude information such as user ID, first name, last name, and username from the results. The front-end will combine these into a single "Guests" score. It will still display the number of students.

I've also updated the setting dialog to be more responsive.

src/components/my-widgets-scores.jsx
```diff
@@ -7,7 +7,7 @@ import LoadingIcon from './loading-icon'
 import NoScoreContent from'./no-score-content'
 import './my-widgets-scores.scss'
 
-const MyWidgetsScores = ({inst, beardMode}) => {
+const MyWidgetsScores = ({inst, beardMode, is_student}) => {
```
clpetersonucf, line 10: It looks like this prop is passed in but never used.

src/components/my-widgets-scores.scss
```diff
@@ -10,6 +10,12 @@
 	overflow-y: auto;
 	padding-bottom: 1em;
 
+	&.limited_because_student {
```
clpetersonucf, line 13: I'm assuming this was related to the `is_student` prop you're passing to `MyWidgetsScores`, but the wires didn't quite get hooked up.

clpetersonucf (changes requested):
Some small things here and there that caught my eye.

Additionally, there appear to be some issues with rendering the settings dialog related to the responsive styles you added, at least on my end:

Screenshot 2024-01-23 at 2 44 25 PM

cayb0rg:
Thanks for spotting these!

> Additionally, there appear to be some issues with rendering the settings dialog related to the responsive styles you added, at least on my end:
> 
> Screenshot 2024-01-23 at 2 44 25 PM

I couldn't replicate this issue (which I assume is the date picker wrapping even though there is room for it). Could you possibly specify the browser, screen width, and/or zoom level?

clpetersonucf:
> I couldn't replicate this issue (which I assume is the date picker wrapping even though there is room for it). Could you possibly specify the browser, screen width, and/or zoom level?

Firefox, 100% zoom, 1792px. I don't recall encountering this issue previously, but it definitely appears related to the addition of `flex-wrap` on line 301 of the settings-dialog stylesheet.

clpetersonucf (approved):
Approved. Really great work here: your solution to score display when a user becomes a student is inspired. It's non-destructive, still gives user score access, but decontextualizes the scores until/unless the student role is replaced.

Well done, I think this is good to go.

PR #1541, opened 2023-10-31
Issue/1508 pagination

### LINKED ISSUE #1508: User and Instance admin search results should be paginated

cayb0rg:
Reopens #1518 
Fixes #1508

* Adds pagination to the admin interface's instance search and admin user search
* Adds offset, limit, and addtional order_by ID to SQL query for instances in My Widgets Page
  * This is also applied to the collaborator and extra attempts dialog
* Adds scrollbar to collaborator dialog user search results
* Set avatar sizes in user admin search to 50x50

~Currently, the user and instance search retrieve all entries in the database on page load. Depending on feedback, this may switch to fetching only after someone has typed something.~ Fixed.


src/components/hooks/useInstanceList.jsx
```diff
@@ -1,9 +1,9 @@
 import { useState, useEffect, useMemo } from 'react'
 import { useInfiniteQuery } from 'react-query'
-import { apiGetWidgetInstances } from '../../util/api'
+import { apiGetWidgetInstances, apiSearchWidgets } from '../../util/api'
 import { iconUrl } from '../../util/icon-url'
 
-export default function useInstanceList() {
+export default function useInstanceList(userOnly = true, query = "") {
```
clpetersonucf, line 6: The more I think about this, the more I want this hook to be split into two different instance list hooks, because while there are some similarities, there are some key differences:

- They're hitting different API endpoints
- They're both requesting lists of widgets, but for different purposes. This especially affects the `widgets` query cache: one is a list of instances the user owns, the other is a list of instances the user may not expressly have access to. IMO, the `widgets` query cache should remain the list of widgets we own (in the My Widgets page) and the new hook would use a different query key.
- It makes references to the hooks elsewhere much cleaner as to what exactly the hook is being used for

clpetersonucf (changes requested):
Good work so far :+1:, some initial feedback items as a first pass

clpetersonucf:
Additional feedback: loading the entire list of users and instances on admin page load is a definite no-go. That's a very expensive request and there's no good reason for it. I'd much rather exclusively return the results of search queries instead.

cayb0rg:
New changes: updated the function names to use "instances" instead of "widgets", created `useSearchInstances` hook from `useInstanceList`, and prevents searching on page load in admin search panel.

clpetersonucf (changes requested):
The only thing I'm seeing is the user panel continues to fetch the entire table's contents on load, which I'm not a fan of. It looks like the infinite query is missing the `enabled: query.length > 0` param present in the instance search query.

The limit in the instance manager's `get_all` method is fine, generally, but I'm not sure when it would realistically (or should realistically) be used. Prior to the reach branch, superusers would have implicit ownership of every instance the database, and thus the API would try and respond with the entire table when visiting My Widgets (because the request wasn't paginated). That request actually caused PHP-FPM to run out of memory. I think I'm fine leaving it in, but it's very much a "you probably shouldn't ever request every instance in the database in one go" situation.

Other than that: LGTM 💪

cayb0rg:
That's a good point; functions querying the widget_instance table are either fetching a page or only the first instance in the results, so it's safe to leave the default limit at 80. 

Caught an issue with paginated results returning duplicate entries because of the `ORDER BY` keyword; removing this seems to have fixed it.

fuel/app/classes/materia/widget/instance/manager.php
```diff
@@ -76,22 +77,23 @@ public static function get_all_for_user($user_id, $load_qset=false)
...
-		// inst_ids corresponds to a single page's worth of instances
-		$displayable_inst = array_slice($displayable_inst, $offset, $widgets_per_page);
+		$items_per_page = 80;
+		$offset = $items_per_page * $page_number;
+
+		// query DB for only a single page of instances + 1
+		$displayable_items = self::get_all($inst_ids, false, false, false, $offset, $items_per_page);
+
+		// if the returned number of instances is greater than a page, there's more pages
+		$has_next_page = sizeof($displayable_items) > ($items_per_page - 1) ? true : false;
```
clpetersonucf, line 91: This is getting into the weeds a bit, but this method of determining whether there's another page will still return true if the total item count is divisible by `items_per_page`.

A better option might be to set the query limit to `$limit + 1`, and simply check whether `$displayable_items` is greater than `$items_per_page`. That shouldn't affect the offset, since the +1 isn't part of the calculation.

fuel/app/classes/materia/widget/instance/manager.php
```diff
@@ -135,20 +137,51 @@ public static function lock($inst_id)
...
+	public static function get_paginated_instance_search(string $input, $page_number = 0)
+	{
+		$items_per_page = 80;
+		$offset = $items_per_page * $page_number;
+
+		// query DB for only a single page of instances + 1
+		$displayable_items = self::get_widget_instance_search($input, $offset, $items_per_page);
+
+		// if the returned number of instances is greater than a page, there's more pages
+		$has_next_page = sizeof($displayable_items) > ($items_per_page - 1) ? true : false;
```
clpetersonucf, line 157: Same thought process as my comment on `instance/manager` line 91.

fuel/app/classes/materia/api/v1.php
```diff
@@ -886,23 +886,37 @@ static public function semester_date_ranges_get()
...
-		$user_arrays = [];
+		$items_per_page = 50;
+		$offset = $items_per_page * $page_number;
 
-		// scrub the user models with to_array
-		if (count($user_objects))
+		// query DB for only a single page + 1 item
+		$displayable_items = \Model_User::find_by_name_search($input, $offset, $items_per_page);
+
+		$has_next_page = sizeof($displayable_items) > ($items_per_page - 1) ? true : false;
```
clpetersonucf, line 906: Same feedback here as my comment in `instance/manager` line 91.

clpetersonucf (changes requested):
Three small requests to change the manner in which you're calculating `$has_next_page`; other than that it's good!

fuel/app/classes/materia/widget/instance/manager.php
clpetersonucf: Update: I'm also wondering if the behavior you noticed with `order by` was influenced by the +1 offset. I'd prefer to keep the `order by` clause when performing the query, so it'd be worth double-checking after updating these.

cayb0rg: Having a single `order_by` seems to have been causing the issue, since multiple instances can have the same `created_at` value. I added an additional criteria to break up ties by ordering by ID, and reinstated the original pagination logic (adding +1 to the limit and then popping the last entry).

clpetersonucf (approved):
The second `order by` clause addition is a nice touch. Nicely done.

This looks good, I think it's ready to ship.

PR #1544, opened 2023-11-14
improves play.auth string check in user admin panel

### LINKED ISSUE #1543: User admin instances played not correctly displaying LTI context
When reviewing instances played by a user in the user admin page, we include a property called "Context" that's supposed to be `Web` for non-LTI plays and `LTI` for LTI plays. However, the string match is case sensitive:

```
{play.auth == 'LTI' ? 'LTI' : 'Web' }
```

And at least on production, the `auth` value is provided as `"lti"` instead of `"LTI"`

clpetersonucf:
Resolves #1543


FrenjaminBanklin (approved):
Checking the type is probably safer than casting `play.auth` to a string in all cases.

Looks good to me.

PR #1551, opened 2023-12-15
Replace Node Sass with Dart Sass

cayb0rg:
Node Sass has been deprecated since October 26, 2020, and it is recommended to use Dart Sass instead. Node Sass also seems to be causing failure during the `app yarn_stage` when creating the docker image on my device.


clpetersonucf (approved):
:+1:

PR #1554, opened 2024-01-08
Fixes to My Widgets score summary user search

### LINKED ISSUE #1550: Individual student search in My Widgets is busted
To reproduce:

1. Visit My Widgets and select an instance with scores
2. Select the Individual Scores tab
3. Enter text in the search input

For every instance I've looked at, searching causes the throbber to appear but the component is otherwise inoperable. There are no console errors reported.

clpetersonucf:
Resolves #1550

- Implements our own custom `useDebounce` hook to un-brick the student search feature
- Adds conditionally rendered "No users found" state when user search returns an empty set
- Two small unrelated fixes while I was digging through associated files: a comment typo and a duplicate, unused class attribute


cayb0rg (approved):
Nicely done!

This is irrelevant to the issue, but while we're looking at the individual scores, we could remove the overlap here:
image

PR #1556, opened 2024-01-09
Instance list request modifications to fix settings persistence

### LINKED ISSUE #1553: Widget settings not properly updated in instance list query cache
To reproduce:

1. Have two widgets in My Widgets
2. Select Widget A and update value(s) in the settings dialog, like Open/Close dates. Hit Save.
3. Select Widget B.
4. Select Widget A again. The settings will represent the original values pulled from the server, no the values previously set.

It should be noted that the correct values are sent to the server and stored in the DB, this is exclusively a display error with the local query cache/state object representing the instance list.

clpetersonucf:
Resolves #1553 

Even though settings are correctly saved to the server and subsequent pulls from the API include the correct data, the local query cache containing the instance list wasn't being updated because the function that converts the raw query data into the actual instance list is memoized for performance. The `useMemo` hook only does a shallow comparison of objects and would not register the changes applied to an individual instance.

Instead, this PR adds a `modified` value to the local query cache when it is altered locally via `setQueryData` - for example, updating, copying, or deleting an instance. This is recognized by the `useMemo` hook which refreshes the instance list appropriately.

Additionally, the `useUpdateWidget` hook was stringifying certain values that should be integers - as far as I can tell there's no need for them to be strings, when the values from the server are always ints. They've been updated accordingly.

### DIFF
--- a/fuel/app/classes/materia/widget/instance/manager.php
+++ b/fuel/app/classes/materia/widget/instance/manager.php
@@ -90,6 +90,7 @@ class Widget_Instance_Manager
 
 		$data = [
 			'pagination' => $displayable_inst,
+			'modified'   => time(),
 		];
 		
 		if ($has_next_page) $data['next_page'] = $page_number + 1;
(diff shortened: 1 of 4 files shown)


cayb0rg (approved):
Nice work! Not seeing anything breaking here 👍🏻

PR #1559, opened 2024-01-11
Fix to player component scoreScreenURL handling

### LINKED ISSUE #1558: Some LTI scores are not passed back due to score screen URL race condition
Score passback is not occurring in the following edge case:
1. Students are playing widgets in an LTI context
2. Students complete a widget and then select Play Again to begin a new session immediately afterwards
3. The `token` GET parameter, assigned at the start of the subsequent play, is sometimes missing from the score screen URL the students are navigated to. The LTI event hook in FuelPHP that looks for score updates doesn't see the `token` parameter, and thus doesn't send anything back to the LTI provider.

We put a lot of work into the widget player to ensure logs are immutable, queued properly, and transmitted properly before a play is finished. If played in an LTI context, a `score_url` property is included in the response from `play_logs_save` when the end log is received. However, score screen URLs are saved on a different state object, and under certain circumstances the `scoreScreenURL` state value is not updated before `window.location.assign(scoreScreenURL)` is called.

clpetersonucf:
Resolves #1558. This is a high priority issue.

The `widget-player` component assigns a default value to the `scoreScreenURL` state object, but if played in an LTI context, the `on_play_completed_event` returns a `score_url` value that's appended to the response object of `play_logs_save` if the score module reports the play is completed.

The problem is that the `scoreScreenURL` state object is updated independently of the rest of the component's state values that manage the overall player state. As best I can tell, under certain circumstances, the `scoreScreenURL` is not updated before the `playState` state value, and the hook responsible for navigating to the score screen fires with the old `scoreScreenURL` value instead. Because of the nature of the race condition, this is very difficult to test.

This PR:
~~Defers the `queueProcessing` flag from being reset to `false` if the `score_url` value is present in the response until after the `scoreScreenURL` is properly updated (via another `useEffect` hook). The `queueProcessing` flag must be false before the `playState` is updated to `end`, which is one of the trigger conditions for navigating to the score screen~~

Replaces the `scoreScreenURL` state object with a `useRef` hook, which is updated immediately and maintained across renders.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,11 @@ cd Materia/docker
 ./run_first.sh
 ```
 
-In a separate terminal window, run `yarn dev` to enable the webpack dev server and live reloading while making changes to JS and CSS assets. Materia is configured to run at `https://127.0.0.1` by default.
+The `run_first.sh` script only has to be run once for initial setup. Afterwards, your local copy will persist in a docker volume unless you explicitly use `docker-compose down` or delete the volume manually.
+
+Use `docker-compose up` to run your local instance. The compose process must persist to keep the application alive. Materia is configured to run at `https://127.0.0.1` by default.
+
+In a separate terminal window, run `yarn dev` to enable the webpack dev server and live reloading while making changes to JS and CSS assets. 
 
 Note that Materia uses a self-signed certificate to facilitate https traffic locally. Your browser may require security exceptions for both `127.0.0.1:443` and `127.0.0.1:8008`.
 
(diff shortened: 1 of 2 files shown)


cayb0rg (approved):
Reproduced the issue using `setTimeout(() => {
							setScoreScreenURL(result.score_url)
						}, 5000)`. This seems to have fixed it. Only potential, and similar, issue I see being is `setQueueProcessing(true)` taking too long, and either nothing happens or duplicate `play_log_save` calls are made, but as I couldn't actually cause this to happen without using `setTimeout`, I don't think it's something to worry about.

clpetersonucf:
In my rush to engineer this fix yesterday, I completely overlooked `useRef` as an alternative to maintaining `scoreScreenURL` in state. Reworked the fix to leverage `useRef` instead, which should be much simpler and reduces the odds of additional race conditions.

clpetersonucf:
Closing this and re-opening as a more comprehensive set of fixes associated with the passback bug.

PR #1560, opened 2024-01-17
Triage fixes for LTI score passback bug

clpetersonucf:
The bug:

As of `v10.0.1-rc.1` and a database running MySQL 8, there appears to be a bug associated with the `DB` session driver where values written to user sessions don't persist. Specifically, the `lti-link-<play_id>` values written when a user replays an instance in an LTI context goes missing. This has multiple downstream impacts, most importantly that upon submission, a replay isn't considered an LTI play and the score is not returned to the LTI provider. It's not yet clear if the underlying cause is related to:

- Changes between MySQL 5.7 and 8 that cause session writes to sometimes go bad
- Issues with session rotation, specifically when the Session table is under significant load
- Issues with FuelPHP 1.9 AND MySQL 8 or some combination of the above

Until the core issue can be identified, this PR serves as a stopgap measure. Changes include:

- Modifies `scoreScreenUrl` in the `widget-player` component to use `useRef` instead of state. While this didn't fix the core issue, it potentially alleviates a race condition.
- Adds `auth` and `environment_data` values to the `\Materia\Session\Play` instance properties when requested via `get_by_id`. These are were not previously included in the DB query.
- Adds "triage" fallback logic to the `ltievents` class when requesting `lti-link-<play_id>` from session. If the token value is not present in session, it interrogates (or creates and then interrogates) a play instance for the `auth` and `environment_data` properties. The play's environment data will include the `token` value that was supposed to be stored in session.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,11 @@ cd Materia/docker
 ./run_first.sh
 ```
 
-In a separate terminal window, run `yarn dev` to enable the webpack dev server and live reloading while making changes to JS and CSS assets. Materia is configured to run at `https://127.0.0.1` by default.
+The `run_first.sh` script only has to be run once for initial setup. Afterwards, your local copy will persist in a docker volume unless you explicitly use `docker-compose down` or delete the volume manually.
+
+Use `docker-compose up` to run your local instance. The compose process must persist to keep the application alive. Materia is configured to run at `https://127.0.0.1` by default.
+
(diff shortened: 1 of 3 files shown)


cayb0rg (approved):
Works as expected on localhost (including when I make the token null on line 353 of `ltievents.php`) and QA (embedded in assignment page and in a new tab). I _think_ this is good to go, but I'm no expert here.

PR #1564, opened 2024-02-06
Fix settings dialog

cayb0rg:
* Fixes grandfathered-in non-owner students from removing their privileges to an instance upon hitting save
* Fixes students from being able to change access levels (depending on whether they're the owner or not)


clpetersonucf (approved):
Good catch, good changes.

PR #1575, opened 2024-04-04
Custom score screen qset acquisition changes

clpetersonucf:
Previously, the qset sent to the score screen was bundled with instance information as part of the `widget_instances_get` API call. Unfortunately this only ever provided the newest copy of a qset, which poses a problem when reviewing older plays that used older copies of a qset, since the IDs for individual items would not match (if they existed at all). 

This PR instead attaches the qset to the response object of `widget_instance_play_scores_get`, which is called when the score screen requests detailed information for a given play. Since the response object already contains a pair of properties, `overview` and `details`, the hope is it makes intuitive sense to attach the `qset` property adjacent to those. The qset provided will match the qset used in each play.

In addition to updating the score manager's `get_play_details` method with the above, the additional parameter for `widget_instances_get` and the associated changes made to support it have been removed.

### DIFF
--- a/fuel/app/classes/materia/api/v1.php
+++ b/fuel/app/classes/materia/api/v1.php
@@ -42,18 +42,18 @@ class Api_V1
 		return Widget_Manager::get_widgets([], $type);
 	}
 
-	static public function widget_instances_get($inst_ids = null, bool $deleted = false, $load_qset = false)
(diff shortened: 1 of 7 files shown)


cayb0rg (approved):
Tested with various saves, works great! I see no issues here.

PR #1580, opened 2024-04-30
Save History fix for new instances

clpetersonucf:
A couple spots in the creator component were using references to the `instId` prop, which is `undefined` or `null` when a new instance is initialized. The value is never updated retroactively, so hooks that prompt the creator to reload via Save History would initialize another new instance instead of using the `initExistingWidget` path.

Updated a couple references to the `instId` prop to reference `instIdRef`'s `current` value instead, and ensured that ref value is updated after the instance is saved (and a fresh instance ID is provided).


cayb0rg (approved):
Not seeing any issues here, and save history now works with new instances. Nice work!

PR #1581, opened 2024-05-01
Support users can view instance scores

### LINKED ISSUE #1457: [Reach Branch] Support users should be able to view standalone score screens associated with a specific play/context ID
Right now, you can review a user's instance play history as a support user, and even have the option to view the score screen in question by selecting the score value in the user admin panel:

Screenshot 2023-03-08 at 1 50 23 PM

However, doing so leads to the score screen in an error state. It appears to be due to the fact that the score manager's `get_instance_score_history` method is searching for instance scores implicitly associated with the current user:
```
$query = \DB::select('id','created_at','percent')
			->from('log_play')
			->where('is_complete', '1')
			->where('user_id', \Model_User::find_current_id())
			->where('inst_id', $inst_id)
			->order_by('created_at', 'DESC');
```

We may want to consider changing this behavior slightly to allow either one of the following cases:
1. Get instance score history for a particular instance for the current user and optionally, a given context ID
2. Get instance score history for a a particular instance if the user is a support user and a context ID is provided

At least right now, the user admin and instance admin panels don't display the full score history of an instance, so it's up for debate whether option 2 should even require a context ID if the user is a support user.

clpetersonucf:
Resolves #1457 (related to Support Users viewing scores)

- Implements the `MyWidgetsScores` and child components in the instance admin panel, allowing a support or super user to review all score details associated with a given instance. No changes were needed to allow support users access to play logs they do not have explicit access to.

- Fixes a bug where directly accessing an instance from the instance admin URL + widget hash was no longer working.

- CSS overhaul of score summary table to make it more responsive, previously it would break visually when viewed in the instance manager embedded within user admin.

### DIFF
--- a/src/components/my-widgets-score-semester.jsx
+++ b/src/components/my-widgets-score-semester.jsx
@@ -93,12 +93,16 @@ const MyWidgetScoreSemester = ({semester, instId, hasScores}) => {
 
 	return (
 		<div className='scoreWrapper'>
-			<h3 className='view'>{semester.term} {semester.year}</h3>
-			<ul className='choices'>
-				{ standardTabElementsRender }
-				{ storageTabRender }
-			</ul>
-			{ activeTab }
+			<section className='scores-header'>
+				<h3 className='semester-label'>{semester.term} {semester.year}</h3>
+				<ul className='choices'>
+					{ standardTabElementsRender }
+					{ storageTabRender }
+				</ul>
+			</section>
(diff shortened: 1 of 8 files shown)


cayb0rg (approved):
Works for super users, support users with student roles, and support users with staff roles. Apart from the merge conflict, LGTM!

cayb0rg (approved):
:ok_hand:

PR #1584, opened 2024-05-08
Prevents play score percentages over 100

### LINKED ISSUE #1392: The percent field in log_play shouldn't accept values over 100
I'm fairly certain there's no case where this should happen, and values > 100 break various aspects of My Widgets' score display.

clpetersonucf:
Resolves #1392 

This is a single-line addition to the `Session_Play` class to ensure the `$percent` param submitted to `set_complete` is never allowed to be over 100. While the database will happily accept values over 100, certain places downstream will have major issues with play percentages that are out of bounds.

---
  - https://app.asana.com/0/0/1207543795708273

### DIFF
--- a/fuel/app/classes/materia/score/manager.php
+++ b/fuel/app/classes/materia/score/manager.php
@@ -152,6 +152,9 @@ class Score_Manager
 		$semesters = [];
 		foreach ($result as $log)
 		{
+			// remove any erroneous > 100% scores from the distribution graph
+			if ($log['bracket'] > 9) continue;
+
 			$key = $log['id'];
 			if ( ! isset($semesters[$key]))
 			{
--- a/fuel/app/classes/materia/session/play.php
+++ b/fuel/app/classes/materia/session/play.php
@@ -368,6 +368,9 @@ class Session_Play
 
 	public function set_complete($score, $possible, $percent)
 	{
+		// ensure percent is never inadvertently set to a value over 100, which will break things
+		if ($percent > 100) $percent = 100;
+
 		// set max score to the current score
 		$max_percent = $percent;
 


cayb0rg:
This does a good job of preventing scores from being over 100 for new scores. It's outside the scope of this PR, but I wonder if it'd be possible to also add a validation check for existing log_play percentages. For existing ones > 100%, I get this error (after merging in dev/10.2.0 - before, the page would just crash):

image

I believe it has to do with `get_widget_score_distribution` putting scores > 100% in their own bracket above 9, which causes `apiGetScoreSummary` to error.

Also, on the scorescreen, before the single-line addition, scores would show as > 100% in the previous scores. Now they're 100%, as expected. However, it will still show as > 100% in the overview table:

image

Just leaving this as a comment as these issues aren't exactly what the PR is addressing.

clpetersonucf:
> Also, on the scorescreen, before the single-line addition, scores would show as > 100% in the previous scores. Now they're 100%, as expected. However, it will still show as > 100% in the overview table

Ideally this will never happen with new score screens, because the percentage saved to the DB will always be 100 or less. Fixing existing records in the `log_play` table with `perc` values over 100 isn't within the bounds of this PR though.

Added additional check to the `get_widget_score_distribution` endpoint to ensure the provided distribution doesn't include brackets outside the intended 0-9 range. This prevents the score summary graph from erroring out in cases where a 200% score may be present.

cayb0rg (approved):
> Ideally this will never happen with new score screens, because the percentage saved to the DB will always be 100 or less. Fixing existing records in the log_play table with perc values over 100 isn't within the bounds of this PR though.

Understandable. May be possible to include a >100 check on the front-end, but also wouldn't make much sense since users who have >100 scores have already seen them lol

Nice, efficient fix to the score distribution view. This is looking good to me 👍🏻

PR #1585, opened 2024-05-17
Fixes/improves expired preview log handling

clpetersonucf:
One overlooked change to the API error state improvements is the way expired preview logs are handled; they no longer tripped the Preview Expired view rendering because the API was returning a `204` instead of an error code. This small PR adds a new `expired()` static method to the Materia `Msg` class, which returns a `410` response (Gone) instead of a `204`. Note that the associated react query has `retry: false` applied to reduce the delay between error receipt and the Expired Preview view being rendered.

---
  - https://app.asana.com/0/0/1207533547776226


cayb0rg (approved):
Great catch! I see no issues with this.

PR #1588, opened 2024-06-27
Switch Asset Storage to S3

cayb0rg:
Switches storage driver to s3. This uses fakes3 on dev and AWS S3 bucket on prod. (**Tested with personal S3 bucket. Not tested on materia server yet**)

Things to note:
* I didn't see the purpose of setting AWS keys in the docker compose file since these are set in .env.local and converted to `$_ENV` vars in `bootstrap.php` during local development, and production also doesn't appear to need them. If this is a mistake, lmk
* This updates the AWS SDK for PHP package, so you might have to do a fresh install (re-run `run_first.sh`)
* The fakes3 container was restarted after running `run_tests.sh`, essentially deleting all uploaded assets. My initial fix for this is just create a whole new fakes3 container for testing, and drop it after tests are complete.

How to test:
* `run_first.sh` should run without errors, meaning all widget assets were uploaded. Check `s3mnt/` folder in local fakes3 container
* Do all thumbnails and uploaded assets load properly in different views?
* Does the media uploader work?
* Does running `run_tests.sh` remove any assets? (Do assets in widget demos still load?)
* Any other way you can think to break it!

If you want to test your own S3 bucket, you'll need to set these variables in your `.env.local` file in the root folder. Here's the [guide](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/getting-started_index.html) I used for setting up AWS locally
```
FAKES3_DISABLED = true
# ASSET_STORAGE_S3_BUCKET=<MUST_SET>
# ASSET_STORAGE_S3_ENDPOINT=<MUST_SET>
# AWS_ACCESS_KEY_ID=<MUST_SET>
# AWS_SECRET_ACCESS_KEY=<MUST_SET>
# AWS_SESSION_TOKEN=<MUST_SET> // STS token for s3 development
```

To Do:
- [x] Add S3 Object Locking in `lock_for_processing` and `unlock_for_processing` for when assets are resized
- [x] Disable Object Locking when using fakes3
- [x] May need a re-check / error handling refactor after dev/10.2.0 is pushed
- [x] Test deleting assets

---
  - https://app.asana.com/0/0/1207667770721396


fuel/app/classes/materia/widget/asset/manager.php
```diff
@@ -121,6 +123,13 @@ static public function delete_asset($id)
 		if (Widget_Asset_Manager::can_asset_be_deleted($id))
 		{
 			$asset = Widget_Asset::fetch_by_id($id);
+			// Uncomment to delete from S3
```
clpetersonucf, line 126: We can probably leave this in, but all the asset manager deletion stuff appears to be vestigial - it's not actually used anywhere, as best I can tell. Simply removing the row from the db is safer, and the one place that actually "deletes" assets (in the `api/asset` controller) doesn't touch this code at all.

fuel/app/classes/materia/widget/asset/storage/s3.php
```diff
@@ -20,14 +20,117 @@ public static function instance(array $config): Widget_Asset_Storage_Driver
...
+				'Retention' => [
+					'Mode' => 'GOVERNANCE',
+					'RetainUntilDate' => new \DateTime('+1 hour'),
+				],
+				// 'VersionId' => '<string>',
+				// 'BypassGovernanceRetention' => true || false,
+				// 'ChecksumAlgorithm' => 'CRC32|CRC32C|SHA1|SHA256',
+				// 'ContentMD5' => '<string>',
+				// 'ExpectedBucketOwner' => '<string>',
+				// 'RequestPayer' => 'requester',
```
clpetersonucf, line 46: Are these values still needed?

fuel/app/classes/materia/widget/asset/storage/s3.php
```diff
@@ -20,14 +20,117 @@ public static function instance(array $config): Widget_Asset_Storage_Driver
...
+				// 'VersionId' => '<string>',
+				// 'RequestPayer' => 'requester',
+			]);
+			// Result syntax:
+			// [
+			// 	'Retention' => [
+			// 		'Mode' => 'GOVERNANCE|COMPLIANCE',
+			// 		'RetainUntilDate' => <DateTime>,
+			// 	],
+			// ]
```
clpetersonucf, line 86: If we're leaving commented-out blocks of code in, they should be accompanied by something that explains their purpose.

fuel/app/config/development/materia.php
```diff
@@ -23,12 +23,31 @@
 
 	// Storage driver can be overridden from env here
 	// s3 uses fakes3 on dev
-	'asset_storage_driver' => 'file',
+	'asset_storage_driver' => $_ENV['ASSET_STORAGE_DRIVER'] ?? 's3',
```
clpetersonucf, line 26: Default value should probably be `file`

fuel/app/config/materia.php
```diff
@@ -96,7 +96,7 @@
 	'google_tracking_id' => $_ENV['GOOGLE_ANALYTICS_ID'] ?? false,
 
 	// Asset storage configuration
-	'asset_storage_driver' => $_ENV['ASSET_STORAGE_DRIVER'] ?? 'file',
+	'asset_storage_driver' => $_ENV['ASSET_STORAGE_DRIVER'] ?? 's3',
```
clpetersonucf, line 99: Same as dev config, default value should probably be `file`

clpetersonucf (changes requested):
Overall this is great. Just a few nit-picks about how asset configuration is setup out of the box. I think `file` is a safer option for most users, as long as we have sufficient documentation to allow new users to swap to `s3` in dev easily.

docker/docker-compose.yml
```diff
@@ -22,11 +22,11 @@ services:
       dockerfile: materia-app.Dockerfile
     environment:
       # View Materia README for env settings
-      - ASSET_STORAGE_DRIVER=file
-      - ASSET_STORAGE_S3_BUCKET=fake_bucket
-      - ASSET_STORAGE_S3_ENDPOINT=http://fakes3:10001
-      - ASSET_STORAGE_S3_KEY=KEY
-      - ASSET_STORAGE_S3_SECRET=SECRET
+      # - ASSET_STORAGE_DRIVER=${ASSET_STORAGE_DRIVER}
```
clpetersonucf, line 25: Storage driver value should probably remain in.

README.md
```diff
@@ -89,4 +89,13 @@ Materia supports two forms of authentication:
...
 
-Materia enables users to upload media assets for their widgets, including images and audio. There are two asset storage drivers available out of the box: `file` and `db`. `file` is the default asset storage driver, which can be explicitly set via the `ASSET_STORAGE_DRIVER` environment variable.
\ No newline at end of file
+Materia enables users to upload media assets for their widgets, including images and audio. There are three asset storage drivers available out of the box: `s3`, `file` and `db`. `s3` is the default asset storage driver, which can be explicitly set via the `ASSET_STORAGE_DRIVER` environment variable.
+
+### Local Asset Storage
+
+By default, a fake `s3` server will be spun up. To test Materia with AWS S3, set the following variables in `.env.local`:
+
+1. Set `FAKES3_DISABLED` environment variable to `true`
```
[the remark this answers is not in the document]
clpetersonucf: Actually, follow-up to this: I hadn't fully realized that this environment variable is not set by default, but can be added to manually disable fakes3. That's good stuff, and I think your design can stay as-is.

clpetersonucf:
One other small issue I'm discovering: the `image/driver.php` class is complaining that uploaded images' width and height are null: `Width and height cannot both be null.` when uploading a new asset through the uploader locally with fakes3.

Is this something you encountered? It appears to prevent thumbs from being created properly.

fuel/app/classes/controller/media.php
```diff
@@ -104,7 +104,15 @@ public function action_upload()
 		];
 
 		$name = Input::post('name', 'New Asset');
-		$asset = Widget_Asset_Manager::new_asset_from_file($name, $file_info);
+
+		try {
+			$asset = Widget_Asset_Manager::new_asset_from_file($name, $file_info);
+		}
+		catch (\Exception $e) {
+			$res->body('{"error":{"code":"15","message":"'.$e->getMessage().'"}}');
```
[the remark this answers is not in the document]
cayb0rg: It's a bitmask value for `E_ERROR | E_WARNING | E_PARSE | E_NOTICE` (https://www.php.net/manual/en/errorfunc.constants.php), as checked on line 34 of the `PhpErrorException` class. However, upon further inspection it seems in `exception_handler()` that the code is also being used to get the severity level, and 15 would be `E_USER_DEPRECATED` (which is wrong, oops), so I guess that's why code 16 is being used on line 121 of `media.php`? Honestly, I feel like code `1` would make more sense here; it'd pass the bitmask for error reporting levels and return the `E_ERROR` severity level. Good catch!

fuel/app/classes/materia/widget/asset/storage/s3.php
cayb0rg: No. I left them commented just in case future developers wanted to know the other possible parameters in this version of s3, but I suppose they could just look at the docs.

clpetersonucf (approved):
I'm happy with how prior feedback was addressed, and S3 now appears fully operational on our dev instance on AWS after additional testing and changes.

I'd like additional validation from others, but for my part, this is good to go.

PR #1596, opened 2024-07-19
Added retry: false to react queries with error states

clpetersonucf:
Like the title says, adds `retry: false` to all instances of `useQuery` with a provided `onError` parameter. These queries have fallback behavior for error states, but the built-in retry mechanism for react query meant that error states were delayed by as much as 10-15 seconds as react query attempts its default 3 retries before failing. While useful in principle, the delays to error states feels real bad and could cause issues where a user continues to try interacting with the page when the server has already kicked them out.

### DIFF
--- a/src/components/header.jsx
+++ b/src/components/header.jsx
@@ -12,7 +12,8 @@ const Header = ({
 	const { data: verified} = useQuery({
 		queryKey: 'isLoggedIn',
 		queryFn: apiAuthorVerify,
-		staleTime: Infinity
+		staleTime: Infinity,
+		retry: false
 	})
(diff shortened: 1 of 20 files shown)


cayb0rg (approved):
Thanks for fixing this! It appears to cover all queries in authenticated views with onError handlers. `apiGetScoreSummary` in the `ScoreOverview` component, while it doesn't have an onError handler, should probably still have a `retry:false` parameter for sake of harmony (if `scoreSummary` is null/undefined, it simply won't render, and the error is handled by the global error handler in `api.js`). Nice work!

PR #1602, opened 2024-09-05
1600/save draft button

chrissolanilla:
In this PR, I addressed an issue where, after calling `Materia.CreatorCore.cancelSave(message)`, the text on the save draft button would remain as "Saving...," which could cause confusion for users. I added three lines of code to update the button text to "Failed to save" whenever the save action fails.

**Key Changes:**
The save draft button text now updates to "Failed to save" when `Materia.CreatorCore.cancelSave(message)` is triggered.

This provides better feedback to the user, making it clear that the save attempt was unsuccessful.
**Note: **The text will only update on widgets that call `Materia.CreatorCore.cancelSave(message)`. Widgets that do not call this function should be updated when the creator attempts to save an incomplete widget.

**Considerations:**
The text could be further improved to be more user-friendly or detailed, such as `"Error saving, please try again"` or `"Please try again."` If you prefer a more descriptive message, I can update this in a follow-up change.

### DIFF
--- a/src/components/widget-creator.jsx
+++ b/src/components/widget-creator.jsx
@@ -455,6 +455,8 @@ const WidgetCreator = ({instId, widgetId, minHeight='', minWidth=''}) => {
 				})
 
 				setCreatorState({...creatorState, heartbeatEnabled: false})
+                //also update the text on the Save Draft Button
+                setCreatorState({ ...creatorState, saveText: 'Failed to save' })
 			} else {
 				setAlertDialog({
 					enabled: true,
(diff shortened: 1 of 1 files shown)


clpetersonucf (changes requested):
Your solution for adding the 5-second delay flip works, but it adds a lot of duplicate code that could be cleaner.

I would look into adding an additional `useEffect()` to track the `isTimeoutRunning` property of `creatorState`, so the timeout only needs to be defined once (within the hook) and can be cleaned up appropriately. Locations where you defined the timeout would only need to update `creatorState.isTimeoutRunning`.

Make sure all of your code changes use tabs, not spaces.

clpetersonucf (approved):
Works great, looks great. Well done.

Issue #1598, opened 2024-07-24
Client side needs enforcement of instance title length


clpetersonucf:
Right now an instance title that's too lengthy returns a 403 error from `widget_instance_save`, triggering a failure state in the creator

chrissolanilla:
I have made changes to my fork with branch 1600/save-draft-button and opened a PR

clpetersonucf:
Addressed via #1604

clpetersonucf:
Closing, now on prod.

PR #1604, opened 2024-09-06
checks the length of the title before saving(base: dev/10.3.0)

chrissolanilla:
For issue #1598 , if a creator enters in a title with a length of 101 characters or more it will return a 403 error. 

**Changes:**
In the `const save = (instanceName, qset, version = 1) => {` part of the code before it does anything I first get the length of `instanceName` and check if its greater than 100. If it is then I run the built in `setAlertDialog({` function and return early notifying to the creator that the title length is too long.

### DIFF
--- a/src/components/widget-creator.jsx
+++ b/src/components/widget-creator.jsx
@@ -384,18 +384,17 @@ const WidgetCreator = ({instId, widgetId, minHeight='', minWidth=''}) => {
 	}
 
 	const save = (instanceName, qset, version = 1) => {
-        //cancel saving of the widget if title is too long to prevent crashing
-        let titleLength = instanceName.length;
-        if(titleLength>100) {
-            setAlertDialog({
-                enabled: true,
-                title: 'Title too long', //the max length for title in my testing is 100
-                message: 'Title must be less than 100 characters',
-                fatal: false,
-                enableLoginButton: false
-            });
-            return;
-        }
+		//cancel saving of the widget if title is too long to prevent crashing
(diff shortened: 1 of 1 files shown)


src/components/widget-creator.jsx
```diff
@@ -384,6 +384,18 @@ const WidgetCreator = ({instId, widgetId, minHeight='', minWidth=''}) => {
 	}
 
 	const save = (instanceName, qset, version = 1) => {
+        //cancel saving of the widget if title is too long to prevent crashing
+        let titleLength = instanceName.length;
+        if(titleLength>100) {
```
clpetersonucf, line 389: `titleLength` definition is unnecessary, you can reference `instanceName.length` here directly

src/components/widget-creator.jsx
```diff
@@ -384,6 +384,18 @@ const WidgetCreator = ({instId, widgetId, minHeight='', minWidth=''}) => {
...
+            setAlertDialog({
+                enabled: true,
+                title: 'Title too long', //the max length for title in my testing is 100
+                message: 'Title must be less than 100 characters',
+                fatal: false,
+                enableLoginButton: false
+            });
+            return;
```
clpetersonucf, line 397: `return false` instead of `return`

clpetersonucf (changes requested):
Does the job. Just two small nit-picks, plus: Materia uses tabs, not spaces, please convert your indentation to tabs.

clpetersonucf (approved):
LGTM

PR #1606, opened 2024-09-10
1605/replace render

chrissolanilla:
In this pull request I went ahead and updated React to the latest version as well as using replacing the deprecated function `ReactDOM.render()` with the new `createRoot()` function as recommend by the [React docs](https://react.dev/reference/react-dom/client/createRoot) . 

**Details:**
As the [React docs](https://react.dev/reference/react-dom/client/createRoot) mentions, the new way of making root components are to get the element and call `createRoot` to create a React root for displaying content inside a browser DOM element. After creating the root element, React needs it to call `root.render()` to render components inside it. 

A minor null check was needed for `src/components/detail-carousel.jsx` . The order in which react renders components is different now when using `createRoot` so without the null check clicking on a widget from the widget catalog would show a blank white page with an error saying that `_pics` is null in the function `snapToImage`. 

This pull request handles and replaces all the old files in the codebase that used `ReactDOM.render()` with the new `createRoot()` function.


src/components/detail-carousel.jsx
```diff
@@ -248,6 +248,9 @@ const DetailCarousel = ({widget, widgetHeight=''}) => {
 
 	const snapToImage = (fast=false) => {
 		const _pics = picScrollerRef.current
+		if(!_pics){
```
clpetersonucf, line 251: Single line expressions in a function can be condensed to:
```
if ( ! _pics) return false
```

If using `return` to short-circuit a function, prefer `return false`.

clpetersonucf (changes requested):
All affected pages are updated, this works just fine. Nice job.

Two small nit-picks:
1. Your import statements `import {createRoot} from "react-dom/client"` should use single quotes, not double quotes.
2. See the comment I left in `detail-carousel.jsx`.

Once those are corrected, this is ready to merge.

chrissolanilla:
With my latest commit to this branch, I have updated all the pages to use single quotes and I returned false in the `detail-carousel.jsx` in one line.

clpetersonucf:
You updated double -> single quotes for `ReactQueryDevTools` but not for your actual import addition:
```
import {createRoot} from "react-dom/client"
```

I can merge this as soon as you update double -> single quotes for the `react-dom/client` imports.

clpetersonucf (approved):
:shipit:

PR #1610, opened 2024-09-27
1595/instance filtering

chrissolanilla:
In this pull request I have added 3 buttons underneath the My Widget Page in Materia for more advanced filtering. 

# Changes 
I ended up changing the styles of the sidebar to be more flexible(no pun intended) so it makes more sense and so that it can comply with my button additions. 

For the actual filters I made them check boxes with a parent of a clickable div. They change color when clicked or when you press the keyboard when they are focused. When they are focused I gave them a nice blue highlight to the parent div similar to how I did it in connections. 

The three filters I added are `Drafts:On/Off` , `Published:On/Off` and `Has Score:On/Off` 
These filters will filter based on the properties of the instance object for `instance` with properties of `draft` for both buttons `Drafts: On/Off` and `Published: On/Off`. For the filter `Has Score:On/Off`, I made an api call to get the scores instead of looking at the properties of the instance since the `attempts` property seems to always be -1.

### DIFF
--- a/src/components/checkbox-button.jsx
+++ b/src/components/checkbox-button.jsx
@@ -1,0 +1,66 @@
+import React, { useEffect, useState } from 'react';
+import './checkbox-button.scss';
+
+const CheckboxButton = ({ labelOn, labelOff, onChange, reset }) => {
+	const [isChecked, setIsChecked] = useState(false);
+	const [isFocused, setIsFocused] = useState(false);
+
+	useEffect(() => {
+		if (reset) {
+			setIsChecked(false);
+			onChange(false); //to tell our x button to reset
+		}
+	}, [reset, onChange]);
+
+
+	const handleDivClick = () => {
+		const newCheckedState = !isChecked;
+		setIsChecked(newCheckedState);
+		onChange(newCheckedState);
+	};
+
+	const handleKeyPress = (event) => {
+		if (event.key === 'Enter' || event.key === ' ') {
+			event.preventDefault();
+			handleDivClick();
+		}
+	};
+
+	const handleFocus = (event) => {
+		setIsFocused(true);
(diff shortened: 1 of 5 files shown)


clpetersonucf (changes requested):
A few high-level comments:

- You have commits from the unrelated #1602 PR on this branch/PR
- The loading throbber and associated visual changes when widget instances are being requested are broken. These elements would overlay the search bar until instance loading is complete.
- The filter buttons don't really work in their current position below the search bar. I would rather the filter options remain hidden within an element that expands within the sidebar when toggled. I can work with you to figure out how that might look.
- Draft and Published filter options are good. `attempts` has nothing to do with whether an instance possesses scores; it's the setting associated with an instance's attempt limit. It's set via the settings dialog in My Widgets (under Collaborate/Copy/Delete). That said, filtering by instances that possess an attempt limit is a good add.
- Similarly, filtering by instances with open/close dates would be ideal.

I appreciate the effort you made in getting score information for each instance, but there are some significant issues with your approach:

- Memoizing the instance score information is good, but please conform to our use of react query to facilitate API requests in the future
- I am very much not a fan of requesting `apiScoreSummaryGet` for each instance that's received. Users can potentially have hundreds of instances, which results in hundreds of requests to the server happening all at once. Instead, I think we should consider appending some minimal score information to widget instance objects themselves on the server as part of the response to `widget_paginate_user_instances_get`. The least impactful version is probably a score count (via `COUNT` from `log_play` associated with an `inst_id`) but I'd also prefer that value isn't even included unless the API request specifically asks for it (via another parameter in the `widget_paginate_instances_get` call). This would mean potential changes to the  `api_v1`, `widget_instance_manager`, and `widget_instance` classes.

clpetersonucf:
Closing, replaced with #1614

PR #1612, opened 2024-10-01
Fix for score module attempt limit validation

### LINKED ISSUE #1611: Score module attempt validation does not include semester
Attempt limits are specific to context ids _and_ semesters: at least that's the idea. If I play a widget directly in Fall 2023, and then again in Spring 2024, those count against each semester's attempt limit, but fall 2023's attempt should not be deducted from spring 2024's attempt limit.

However, the score module appears to be ignoring semesters when validating the current attempt limit: https://github.com/ucfopen/Materia/blob/master/fuel/app/classes/materia/score/module.php#L99

This means that for a given context ID, multiple semesters' worth of attempts are being bundled together to count against the attempt limit. The student will be able to initialize the widget, but will be unable to complete it, as the server responds with a 403 to `play_logs_save`.

clpetersonucf:
Resolves #1611. Adds the `$semester` param to the `\Materia\Score_Manager::get_instance_score_history()` reference in the score module as part of its attempt limit validation.

### DIFF
--- a/fuel/app/classes/materia/score/module.php
+++ b/fuel/app/classes/materia/score/module.php
@@ -96,7 +96,8 @@ abstract class Score_Module
 			// Check for attempt limit prior to submission, but only for actual plays (not previews)
 			if ($this->play_id != -1)
 			{
-				$attempts_used = count(\Materia\Score_Manager::get_instance_score_history($this->inst->id, $this->play->context_id));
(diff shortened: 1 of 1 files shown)


DynieM (approved):
Tested changes on local and score attempts now reset based on the semester the student is playing the widget.

PR #1614, opened 2024-10-14
1595/instance filtering2

chrissolanilla:
# Instance Filtering

In this PR I added 7 buttons that act as filters in the My Widgets page of Materia. The filters include filtering by if an instance is a Draft, Published, Limited Attempts, Is Open, Is Expired, Guest Access, and Is Embedded. This PR is nearly identical to my 1595/instance-filtering branch but without the unrelated commits. 

While filtering instances based on score data is possible, it may require significant structural changes to optimize performance. Potential changes could include:

-   **Table Indexing**
-   **Aggregate Queries**
-   **Caching Methods (e.g., Redis)**
-   **Creating a New Table for Score Retrieval**

However, these enhancements would go beyond the scope of this pull request.

## Changes
In addition to adding filters, I have also refactored some of the html and css for the existing page to be more maintainable such as getting rid of absolute positioning, random margins, and opted to use flex and grid layouts to achieve the same look and so it could work with my new front end components. 

The main front-end component I added was `checkbox-button.jsx` which is a nice toggle button built using a hidden checkbox and a parent div with it's own custom styles included with focus support and voice-over accessibility.

### DIFF
--- a/src/components/checkbox-button.jsx
+++ b/src/components/checkbox-button.jsx
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import './checkbox-button.scss';
 
-const CheckboxButton = ({ labelOn, labelOff, onChange, reset, customStyle, fontSize }) => {
+const CheckboxButton = ({ labelOn, labelOff, onChange, reset, customStyle, fontSize,ID }) => {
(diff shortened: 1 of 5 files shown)


clpetersonucf:

PR #1615, opened 2024-10-14
Fix #1607: Retire cookie class

### LINKED ISSUE #1607: Retire Cookie core class override
When `SameSite` cookie requirements first became an issue, FuelPHP did not yet support those headers appropriately out of the box. They are now supported within the core Cookie class, which makes our override copy of the class more or less obsolete. It could potentially be a liability as additions are made to the core class that are not reflected in our own copy.

It should be reasonable to retire [our copy of the Cookie class ](https://github.com/ucfopen/Materia/blob/master/fuel/app/classes/materia/fuel/core/cookie.php) in favor of the one in `fuel/core` and add the required configs to set the `SameSite=None` and `Secure=true` cookie headers.

fudgeu:
This PR addresses #1607, which removes the custom cookie class in favor of the default cookie class in FuelPHP.

Changes largely include:
- Removal of the custom cookie class
- Removal of the code telling Fuel to use the custom cookie class
- Changes to `config/config.php` and `config/session.php` to tell FuelPHP to use `Secure` and `SameSite=None` by default

The changes seem to work, as all applicable network responses I've observed seem to come with both attributes set in the cookie, and Firefox never complains about any cookie warnings in the console. Though, a second look-through just to make sure behavior is like before would be appreciated.


clpetersonucf:
Updated target branch to `dev/10.3.0`

fuel/app/config/config.php
```diff
@@ -196,18 +196,20 @@
...
+	'cookie' => array(
 		// Number of seconds before the cookie expires
 		// 'expiration'  => 0,
 		// Restrict the path that the cookie is available to
 		// 'path'        => '/',
 		// Restrict the domain that the cookie is available to
 		// 'domain'      => null,
 		// Only transmit cookies over secure connections
-		// 'secure'      => false,
+		'secure'      => $_SERVER['HTTPS'] ?? false,
```
clpetersonucf, line 207: Unfortunately I don't think relying on `$_SERVER['HTTPS']` is going to work for us, since this relies on the webserver running from the same machine as the PHP application. With the docker setup we have now, the webserver is running in its own container and forwarding requests to the PHP container through docker networking. The next best thing might be yet another env variable, but realistically I'm not sure there would ever be a context where someone wants to run the server application without https enabled. Perhaps setting the default to `true` is okay here, alongside an optional env variable override.

fuel/app/config/config.php
```diff
@@ -196,18 +196,20 @@
...
 		// Only transmit cookies over HTTP, disabling Javascript access
 		// 'http_only'   => false,
-	// ),
+		// Samesite restrictions on cookie, options include Lax, Strict or None.
+		'same_site'   => (isset($_SERVER['HTTPS']) and $_SERVER['HTTPS']) ? 'None' : 'Strict',
```
clpetersonucf, line 211: Same as above - although in the case of `same_site`, I feel like we could just stick to a default, discrete value of `None`.

fudgeu:
Made the following changes:
- Added a new optional environment variable `IS_SERVER_HTTPS`. If unset or has an invalid value, will default to true.
- `secure` now refers to the `IS_SERVER_HTTPS` variable. Again, if the variable is unset, then it is defaulted to true.
- `SameSite` is now always `None`, regardless of if the server is HTTPS or not.

clpetersonucf (approved):
Good stuff. Looks ready to ship. 💪

PR #1617, opened 2024-10-29
Docker image management updates + run_first script improvements

clpetersonucf:
This spiral'd out after first noticing that the images in the base compose file don't actually exist, causing `run_first` to always build from source instead of pulling pre-built images for `app` and `webserver`.

Updates include:
- Interactive prompt at the init of `run_first` to ask whether you want to pull precompiled images or build them from source
- Removed some redundant steps from the `run_first` script
- Renamed the `app`, `webserver`, and `fakes3` images in the base compose file to their representative tags in the github package registry. They will default to the latest `app-dev`, `webserver-dev`, and `fake-s3-dev` accordingly.
- Added `app-stable` and `webserver-stable` tags to published images for release builds.
- Finally got around to updating the base `mysql` image from `5.7` to `8.0.32`.

# How to test

You'll need to kill your existing development instance of Materia, if present. Sorry about that.

**If you had an existing Materia instance, do this first:**
```
$ docker compose down
$ docker rmi -f $(docker images -q)
```
Then, in the `docker/` directory of Materia:
```
$ ./run_first.sh
```

Choose option 1 for precompiled images. Choose option 2 to build the images locally.


FrenjaminBanklin (approved):
The modifications to the first run shell script work in both scenarios, and initial setup is way faster now when opting to use the existing images.

Issue #1583, opened 2024-05-08
Improve instance lock management


clpetersonucf:
Instance lock management is a little screwy right now. Currently:

`widget_instance_lock` in the v1 API will lock an instance for a given user for two minutes. `widget_instance_lock` is called once when the creator first initializes, but never again. The cache item associated with the lock method persists for two minutes, which means after two minutes the instance is again considered editable. Additionally, the creator will enter a failure state if trying to edit an instance locked by someone else, but it's not graceful - the error dialog is blank and it can be dismissed without reloading or exiting the page.

Two improvements to this feature could include:
1. Ensure the `widget_instance_lock` request is performed on a two-minute interval while someone is on the creator page (easiest solution might be to set the query's `staleTime` to two minutes).
2. Improve creator failure state when the lock request returns false.

PR #1633, opened 2025-01-14
Fixes instance lock error message, staleTimer, and cacheTimer

chrissolanilla:
This addresses issue #1583 where collaborators of a widget are not meant to edit a widget at the same time. Normally this does not happen, but it was still possible to edit a widget without acquiring the lock for the widget by copy-pasting the link of the creator into the browser.

 When they do this now, they are met with an informative alert and it prevents them from being able to close the alert or edit the widget. Additionally, the staleTimer was set to infinity before, which was responsible for requesting the key to the lock. The original creator would hold the lock for 2 minutes but after that it would not do more requests. This PR aims to fix this by adding 2 minutes to the staleTimer in addition adding 3 minutes for the cacheTimer to get rid of the cache every 3 minutes. 
 
 There was a minor bug on the include.scss where a dark mode style would be applied as well, I also fixed that with Simon's help.

### DIFF
--- a/src/components/include.scss
+++ b/src/components/include.scss
@@ -307,7 +307,7 @@ header {
 	}
 
 	a {
-		color: #0093e7; 
+		color: #0093e7;
 		text-decoration: none;
 
 		&:hover {
@@ -527,13 +527,13 @@ header {
 		height: 16px;
 		margin: 0 0.5em 0 0;
 		padding: 0;
-	
+
 		background-color: #fff;
 		border: 2px solid #7e7e7e;
 		border-radius: 4px;
-		
+
 		transition: all 0.3s;
-	
+
 		&:after {
 			content: '✓';
 			position: absolute;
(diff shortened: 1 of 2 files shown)


clpetersonucf (changes requested):
This sorta works but relying on expiring the query data through `staleTime` and `cacheTime` only works when the query data is referenced - this doesn't happen on any sort of automatic interval in our code, so the only way to reliably get the page to refetch the data is by tabbing away from the window and back.

In addition to `staleTime` and `cacheTime`, I think the param you're looking for is `refetchInterval`. I'd also probably want to set `refetchIntervalInBackground` to `true`.

Make sure you verify the lock mechanism works as expected by keeping the page in the foreground for the interval duration.

clpetersonucf (approved):
`refetchInterval` did the trick :+1:

PR #1639, opened 2025-02-04
10.3.1 style fixes

clpetersonucf:
Miscellaneous style fixes for issues associated/not yet addressed with the 10.3 release:

- Fixed dark mode style leak in catalog filter buttons.
- Addressed issues with sharing draft instances in My Widgets. The Share With Your Students section is now more aggressive about disabling and updating certain information when an instance is a draft.

Not style fixes:

- Resolved an issue with the widget creator component that could cause the qset query to occasionally fail to fire. Ensured widget ID and instance ID (if available) is available to the component immediately and reworked the qset query to rely on a hook instead of the deprecated `onSuccess` and `onFailure` callbacks.

### DIFF
--- a/src/components/catalog.jsx
+++ b/src/components/catalog.jsx
@@ -306,14 +306,12 @@ const Catalog = ({widgets = [], isLoading = true}) => {
 						</div>
 					</div>
 					{ mobileFilterRender }
-					<div id='filters-container'
-						className={`ready ${state.showingFilters ? 'open' : 'closed'}`} aria-hidden={!state.showingFilters}>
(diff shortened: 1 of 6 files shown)


fudgeu:
Code looks good to me, and attempted to test out as much of the changes as possible (hard to test if the creator qset query will fail to fire still, but it has worked every time so far). Don't see any issues!

FrenjaminBanklin (approved):
Noticed the `filters-container` element ID appears on two nodes in the catalog component, but that's a relatively minor issue that's easily fixed.

The rest of this looks good and seems to be working well.

PR #1678, opened 2026-03-20
 Add composite index on log_play to improve query performance 

GoldenLin9:
Fixes https://university-of-central-flori-pk.sentry.io/issues/7251704322/?project=4510750267473920&query=&referrer=issue-stream

- Slow db query caused by multiple indexes being in a query

#Testing
Screenshot 2026-03-20 at 11 03 44 AM

# Resources
- https://dev.mysql.com/doc/refman/8.4/en/multiple-column-indexes.html
- https://dev.mysql.com/doc/refman/8.4/en/descending-indexes.html


clpetersonucf (approved):
Sound improvements, as discussed.

PR #1679, opened 2026-03-20
Adjusts makefile recipes to use hard-coded Python container name.

### LINKED ISSUE #1677: Hard-code the python container name
Originally https://github.com/ucfcdl/Materia/issues/31

> Most of the Makefile changes are predicated on being able to correctly identify an existing Python container assuming it's named after the directory the application is running in, which should be fine but is obviously error-prone.
>
> Ideally the Python container should have a predictable name. This should be a pretty trivial change to the docker-compose.yml configuration to hard-code a container name.
>
>As a bonus, the MWDK feature to install widgets directly into a Docker container would be able to make use of a predictable name to drastically simplify that process as well.

FrenjaminBanklin:
Closes #1677

Adjusts makefile Docker container name expectations to match hard-coded value in `docker-compose.yml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,7 @@
 PYTHON_VERSION="3.12.1"
 COMPOSE_FILE=docker/docker-compose.yml
 DOCKER_COMPOSE=docker compose -f $(COMPOSE_FILE)
-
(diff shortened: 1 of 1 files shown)


clpetersonucf (approved):
No notes.

PR #1685, opened 2026-03-23
Use URL params to determine if a creator is embedded.

### LINKED ISSUE #1673: Use URL parameters to govern 'type' in the creator and player pages.
Currently in the components for both the creator and player pages we determine the context in which the page is loaded (embed, preview, demo, etc.) by checking for specific strings in the URL itself, see [here](https://github.com/ucfopen/Materia/blob/17815afcba58e082ea1a275870e4e190649a4a75/src/components/widget-creator-page.jsx#L7) (creator) and [here](https://github.com/ucfopen/Materia/blob/17815afcba58e082ea1a275870e4e190649a4a75/src/components/widget-player-page.jsx#L6-L23) (player).

This may be accomplished more reliably by condensing the possible paths to the creator and player pages and instead using URL parameters e.g. `materia/player/instid?type=embed` or `materia/player/instid?type=preview` to determine 'type'.

FrenjaminBanklin:
Closes #1673.

Removes window location parsing logic from WidgetCreatorPage component in favor of using the '?is_embedded' URL parameter to determine whether a creator is embedded.

Adds new prop to WidgetCreator component to track whether a creator is embedded and use this to control the widget preview URL.

Currently this is probably only ever going to be used by the MWDK to omit unnecessary header components (see ucfopen/Materia-Widget-Dev-Kit#149).

### DIFF
--- a/src/components/widget-creator-page.jsx
+++ b/src/components/widget-creator-page.jsx
@@ -5,18 +5,17 @@ import WidgetCreator from './widget-creator'
 import './widget-creator-page.scss'
 
 const EMBED = 'embed'
-const PREVIEW_EMBED = 'preview-embed'
 
-const getWidgetType = path => {
+const getWidgetType = () => {
+	const urlParams = new URLSearchParams(window.location.search)
 	switch(true) {
-		case path.includes('/embed/'): return EMBED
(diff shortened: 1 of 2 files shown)


clpetersonucf (approved):
Sensible changes, LGTM.

PR #1686, opened 2026-03-24
Score screen indicates max score in LTI contexts

### LINKED ISSUE #1674: Score screen for LTI plays should indicate max score returned vs present score
Originally mentioned in #1317

Materia sends the highest score percentage for the current context whenever it submits a score to Canvas upon play session completion. For example, if I play a widget twice and receive a 100% followed by an 80%, Materia will send 100% for both attempts.

This is not visible to the student when they are sent to the score screen, creating potential confusion as to why the score screen yields one value but the gradebook yields another. Given that we're already sending some LTI-related information over the wire in the `lti` section of the `/api/scores/details/` response object, it would make sense to include a `max_score` value and add some extra UI to the LTI status section of the score screen to address the discrepancy.

clpetersonucf:
Resolves #1674 

- Adds `score_submitted` to `LtiPlayState` model, with `null=True` to allow null values for previously stored play states.
- The `lti` section of the score details endpoint now includes the `score_submitted` value as `max_score`.
- Adjusted score screen components to check for a `max_score` value and if different from the previous score, the LTI overview section indicates the max score value submitted for the given play:

Screenshot 2026-03-24 at 5 00 33 PM


FrenjaminBanklin (approved):
Everything looks sane enough.

Rubber stamping so we can get this running in a more easily testable environment.

PR #1694, opened 2026-03-30
Fixes S3 asset storage driver crashing on audio uploads.

### LINKED ISSUE #1693: The S3 asset storage driver crashes when uploading audio files.
The S3 storage driver [tries to generate a thumbnail immediately after uploading a new file to S3](https://github.com/ucfopen/Materia/blob/17815afcba58e082ea1a275870e4e190649a4a75/app/storage/s3.py#L181-L183) in order to take advantage of existing resources and prevent extra overhead that would be incurred by creating thumbnails later.

Unfortunately this introduced buggy behavior in the case of uploading audio, as the driver tries to make a thumbnail but can't do so because the file it's trying to use as an image is not an image.

FrenjaminBanklin:
Fixes #1693.

Adds a quick mime type check to the S3 driver so it doesn't crash trying to generate thumbnails for audio files.

### DIFF
--- a/app/storage/s3.py
+++ b/app/storage/s3.py
@@ -178,9 +178,13 @@ class S3AssetStorageDriver:
             Key=S3AssetStorageDriver.get_key_name(asset.id, "original"),
             ExtraArgs={"ContentType": asset.get_mime_type()},
         )
-        # Upload the thumbnail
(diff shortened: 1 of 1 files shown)


clpetersonucf (approved):
No notes, easy fix.

PR #1695, opened 2026-04-01
Keep url parameters in embedded creators after saving a new widget for the first time.

FrenjaminBanklin:
Adds a one-liner to carry forward embedded creator URL parameters after saving a new widget instance for the first time.


clpetersonucf (approved):
Looks good to me.

PR #1698, opened 2026-04-02
Adds missing break statement to general error component

### LINKED ISSUE #1697: Missing break statement in general error component
The `error_launch_recovery` error type never gets displayed because the switch/case always falls through.

Relevant code:

https://github.com/ucfopen/Materia/blob/17815afcba58e082ea1a275870e4e190649a4a75/src/components/lti/error-general.jsx#L75-L84

Fix: Add missing `break;` statement.

alexanderpaolini:
Fixes #1697 

As per title. There was a missing break statement from changes a few weeks ago. This PR adds in that break statement.


FrenjaminBanklin (approved):
Can't argue with that.

clpetersonucf (approved):
Good catch. Bonus points for using the correct branch naming scheme and picking the proper target branch :+1:

Issue #1700, opened 2026-04-06
Missing alternative text on many decorative images


alexanderpaolini:
The Materia frontend contains code like `<img src="/static/img/social-ucf-open.png"/>` (`src/components/homepage.jsx:30`), where a decorative image has no alt text. The standard is empty text, rather than no text at all.

clpetersonucf:
Resolved in #1701

PR #1701, opened 2026-04-06
Add null alternative text on decorative images

alexanderpaolini:
As per #1700, many of the decorative images lack alternative text. This PR adds null alt text in many of those places. 

There are a few places I missed. Namely, buttons and iframes that are missing names. Mostly because I don't know enough about the structure of Materia to know what they do and how they should be labeled. For instance, `src/components/guide-page.jsx:38`.


clpetersonucf (approved):
Simple enough, looks good.

PR #1708, opened 2026-04-16
Fixes custom playdata exporters bricking widget install

### LINKED ISSUE #1707: Custom playdata exporters cause widget engine installation to fail
This appears to be because the widget installer is attempting to serialize a `dict_keys` object, which is not possible:

```
manifest_data["meta_data"]["playdata_exporters"] = exporter_mappings.keys()
```

The correction appears to be to simply convert the `dict_keys` to a list:

```
manifest_data["meta_data"]["playdata_exporters"] = list(
    exporter_mappings.keys()
)
```

clpetersonucf:
Resolves #1707

### DIFF
--- a/app/core/services/widget_installer_service.py
+++ b/app/core/services/widget_installer_service.py
@@ -9,13 +9,12 @@ from json import JSONDecodeError
 from pathlib import Path
 
 import urllib3
-from django.utils import timezone
-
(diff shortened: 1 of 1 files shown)


FrenjaminBanklin (approved):
Confirmed, solves the problem.

Issue #1692, opened 2026-03-30
User admin panel: improve playsession request behavior


clpetersonucf:
The user admin panel requests the full set of `/api/play-sessions/?user=12345` records when the page initially loads. This can be a painfully large query depending on the user's interaction history with the platform, and because `include_activity` is included the individual page count is very low (~6 records). Ideally:

1. The records are requested in larger sets for this specific use case,
2. The complete set of play records is not pulled in one go, but rather more like the profile page, where the user chooses to load the next set of records,
3. The manner in which records load in is smoother, ideally using [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) or a similar DOM method to keep the last requested records in view.

clpetersonucf:
Resolved in #1710

PR #1710, opened 2026-04-22
Improves play session loading in user admin panel

GoldenLin9:
Fixes issue #1692

## Problem
The user admin panel auto-fetches all play session records on page load. With include_activity enabled, page size is only 8 records, resulting in many small API requests that are slow for users with large play histories.

 ## Solution
  - Added admin_activity query param that increases page size to 50
  - Replaced auto-fetch with manual "Show more" button
  - Added smooth scrollIntoView to keep new records in view when loading more

### DIFF
--- a/src/util/api.js
+++ b/src/util/api.js
@@ -604,7 +604,8 @@ export const apiCanBePublishedByCurrentUser = (widgetId) => {
 /** Controller_Api_User */
 
 export const apiGetUserPlaySessions = (user, pageParam = 1, admin_activity = false) => {
-	return handleRequest('GET', `/api/play-sessions/?user=${user}&include_activity=true&admin_activity=${admin_activity}&page=${pageParam}`)
+	const activityParam = admin_activity ? 'admin_activity=true' : 'include_activity=true'
+	return handleRequest('GET', `/api/play-sessions/?user=${user}&${activityParam}&page=${pageParam}`)
 }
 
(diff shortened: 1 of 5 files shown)


src/util/api.js
```diff
@@ -603,8 +603,8 @@ export const apiCanBePublishedByCurrentUser = (widgetId) => {
 
 /** Controller_Api_User */
 
-export const apiGetUserPlaySessions = (user, pageParam = 1) => {
-	return handleRequest('GET', `/api/play-sessions/?user=${user}&include_activity=true&page=${pageParam}`)
+export const apiGetUserPlaySessions = (user, pageParam = 1, admin_activity = false) => {
+	return handleRequest('GET', `/api/play-sessions/?user=${user}&include_activity=true&admin_activity=${admin_activity}&page=${pageParam}`)
```
clpetersonucf: One small nit-pick: I'd rather `admin_activity` only be included as a query param when it would be true. It's not that sensitive a value, but given its administrative function, it'd be nice to not include it by default in every request to this endpoint. Similarly, it could replace `include_activity` in situations where it would be applied, instead of being appended in addition to `include_activity`.

clpetersonucf (changes requested):
One really small request associated with the query params - the rest is good.

clpetersonucf (approved):
LGTM

Issue #1691, opened 2026-03-30
Visual issues with closed widgets in embedded contexts


clpetersonucf:
1. HTML entities are not being properly converted in the title text
2. The page is not properly recognizing the embed context and is rendering in a non-embed mode (e.g., the header is visible)

<img width="696" height="432" alt="Image" src="https://github.com/user-attachments/assets/7da2bf02-ebbf-4f3e-8353-10793491542d" />

clpetersonucf:
Resolved in #1713

PR #1713, opened 2026-04-23
Fixes HTML entities displaying as raw text

GoldenLin9:
Fixes issue #1691

## Problem
Django's `{{ value }}` auto-escaping converts `&` to `&amp;`, which is correct for HTML contexts. But inside `<script>` tags the browser does not decode HTML entities, so JavaScript receives the literal string `&amp;` instead of `&`.

## Solution
Replace `{{ value }}` with `{{ value|escapejs }}`. This filter uses JavaScript-native escape sequences that the JS engine properly interprets, while still protecting against code injection.


clpetersonucf (approved):
Easy fix.

PR #1715, opened 2026-04-30
Handles empty semesters param in export_playdata endpoint

GoldenLin9:
Fixes Sentry error: https://university-of-central-flori-pk.sentry.io/issues/7450925333/?project=4510750267473920&query=is%3Aunresolved%20issue.category%3A%5Berror%2Coutage%5D&referrer=issue-stream

## Problem
When `semesters` query param was empty, `"".split(",")` returns `[""]` instead of `[]`, causing the endpoint to crash trying to unpack an empty string into `[year, term]`.

## Testing
Added `test_export_with_empty_semesters_param` to cover this case.

### DIFF
--- a/app/api/tests/test_widget_instance_views.py
+++ b/app/api/tests/test_widget_instance_views.py
@@ -1540,6 +1540,19 @@ class TestInstanceExportPlaydata(WidgetInstanceViewSetTestCase):
 
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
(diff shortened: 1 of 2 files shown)


FrenjaminBanklin (approved):
Stumbled for a few minutes trying to figure out why things were still erroring out before realizing the score and event exporters necessarily should break if valid semesters are not specified.

The fix works as intended for the 'Questions and Answers' export.

clpetersonucf (approved):
:shipit:

PR #1718, opened 2026-05-05
Rework to collaboration dialog

clpetersonucf:
Redesigns the collaboration dialog based on @ToryDeutsch's mockups:

The current user is always positioned at the top of the list, with the UI updated to better communicate their access and how to remove it. Other users are populated in a separate section below. Functionally, the dialog remains the same.

Screenshot 2026-05-05 at 12 26 59 PM

The provisional access notice was slightly tweaked:

Screenshot 2026-05-05 at 12 27 14 PM

Additionally, users with View Scores access no longer see access controls at all:

Screenshot 2026-05-05 at 12 27 39 PM

### DIFF
--- a/src/components/include.scss
+++ b/src/components/include.scss
@@ -2,16 +2,19 @@
 $color-features: #0093e7;
 $color-features-active: #0357a5;
 $color-red: #ca3b3b;
+$color-light-red: #f26666;
 $color-yellow: #ffd439;
 $color-yellow-hover: #ffc904;
 $color-yellow-bg: #ffeca8;
 $color-orange: #ffba5d;
+$color-dark-orange: #864c00;
 $color-green: #c5dd60;
 $color-purple: #b944cc;
 
 $very-light-gray: #ececec;
 $light-gray: #dadada;
 $gray: #9a9a9a;
+$moderate-gray: #797979;
 $extremely-dark-gray: #333;
 
 $color-background-dark: #21232a;
--- a/src/components/my-widgets-collaborate-dialog.jsx
+++ b/src/components/my-widgets-collaborate-dialog.jsx
@@ -257,7 +257,8 @@ const MyWidgetsCollaborateDialog = ({onClose, inst, myPerms, otherUserPerms, set
 					onChange={(e) => setState({...state, searchText: e.target.value})}
 					className='user-add'
 					type='text'
-					placeholder="Enter a Materia user's name or e-mail"/>
+					placeholder="Enter a user's name or e-mail"/>
+				<span className="collab-input-disclaimer">Only individuals who have previously used Materia will show up in search.</span>
 				{ searchResultsRender }
 			</div>
 		)
@@ -273,7 +274,11 @@ const MyWidgetsCollaborateDialog = ({onClose, inst, myPerms, otherUserPerms, set
 		mainContentRender = <NoContentIcon />
 
 		if (containsUser) {
-			const mainContentElements = Array.from(state.updatedAllUserPerms).map(([userId, userPerms]) => {
+
+			const mainContentElements = []
+			let userContentElement = null
+			
+			Array.from(state.updatedAllUserPerms).map(([userId, userPerms]) => {
 				if (userPerms.remove === true) return
 
 				let user = collabUsers[userId]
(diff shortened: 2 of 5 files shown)


src/components/my-widgets-collaborate-dialog.jsx
```diff
@@ -284,22 +289,44 @@ const MyWidgetsCollaborateDialog = ({onClose, inst, myPerms, otherUserPerms, set
...
+							user={user}
+							perms={userPerms}
+							myPerms={myPerms}
+							isCurrentUser={currentUser.id === user.id}
+							onlyOneFullPermHolder={onlyOneFullPermHolder}
+							removedCurrentUser={removedCurrentUser}
+							onChange={(userId, perms) => updatePerms(userId, perms)}
+							readOnly={myPerms?.can?.share === false}
+						/>
+					)
```
alexanderpaolini: I think it would read just a bit better if we did something like (disregard variable names):

```js
const rowElement = <CollaborateUserRow {...} />

if (currentUser.id === user.id) userContentElement = rowElement
else mainContentElements.push(rowElement)
```

src/components/my-widgets-collaborate-dialog.jsx
```diff
@@ -273,7 +274,11 @@ const MyWidgetsCollaborateDialog = ({onClose, inst, myPerms, otherUserPerms, set
 		mainContentRender = <NoContentIcon />
 
 		if (containsUser) {
-			const mainContentElements = Array.from(state.updatedAllUserPerms).map(([userId, userPerms]) => {
+
+			const mainContentElements = []
+			let userContentElement = null
+			
+			Array.from(state.updatedAllUserPerms).map(([userId, userPerms]) => {
```
alexanderpaolini: `.map` can be replaced with `.forEach`, or `for (const [userId, userPerms] of {...})`, since the return value of `.map` isn't being used.

src/components/include.scss
```diff
@@ -2,16 +2,19 @@
...
 $color-orange: #ffba5d;
+$color-dark-orange: #864c00;
 $color-green: #c5dd60;
 $color-purple: #b944cc;
 
 $very-light-gray: #ececec;
 $light-gray: #dadada;
 $gray: #9a9a9a;
+$moderate-gray: #797979;
 $extremely-dark-gray: #333;
```
alexanderpaolini, line 18: Is there any reason why these colors aren't prefixed with `color-`? Or alternatively is there a reason why the other ones are?

src/components/my-widgets-collaborate-dialog.jsx
clpetersonucf: Good suggestions. Updated.

src/components/my-widgets-collaborate-dialog.jsx
clpetersonucf: Good catch! Fixed.

src/components/include.scss
clpetersonucf: No reason, it's arbitrary. You're right that they should all be prepended by `color-`, but addressing that is outside the scope of this PR.

src/components/my-widgets-collaborate-dialog.jsx
```diff
@@ -282,24 +287,31 @@ const MyWidgetsCollaborateDialog = ({onClose, inst, myPerms, otherUserPerms, set
 					return <div key={userId}></div>
 				}
```
alexanderpaolini: This would do the same as 
```js
if (!user) return;
```
I'm not sure why there's a div, but it wasn't being used with the `.map`, and still isn't used for the `.forEach`. I assume its residual code and can be removed

clpetersonucf: Updated.

FrenjaminBanklin (approved):
This looks good and works well.

The only gotcha I could find is if a user somehow has full access that is also provisional, and that user is viewing the collaboration dialog. They'll appear under the 'You' area and the prompt to remove provisional access will still appear, but trying to remove their own provisional access and save permissions yields an error and prevents any other permissions changes from saving.

I don't think it's realistic to expect this scenario outside of forcing it to happen, but it may be worth addressing in a separate future issue. Otherwise this looks like it's good to go.

PR #1720, opened 2026-05-08
Replaces f string with `%s`

### LINKED ISSUE #1717: LtiLaunchMixin recovery error uses f string instead of `%s`
Should be a single-line fix to the error message to use `%s` instead: https://github.com/ucfopen/Materia/blob/master/app/lti/mixins.py#L33

alexanderpaolini:
Replaces f string with `%s` in the logger code for `app/lti/mixins.py`. 

Resolves #1717


clpetersonucf (approved):
Easy

PR #1721, opened 2026-05-08
skips access window check for widget previews

### LINKED ISSUE #1716: Widget previews are blocked by access settings
If I set an access window that opens in the future, and choose to preview a widget, the preview is blocked in the same manner as a play. Previews should always be allowed regardless of open and close date settings.

GoldenLin9:
Fixes #1716

## Summary
Previews were incorrectly blocked by open/close date access window settings the same way a regular play would be. Updated the validation logic to skip the access window check for previews.

## Testing
Manual testing confirmed previews are no longer blocked by access window settings.


clpetersonucf (approved):
🚢

PR #1729, opened 2026-05-13
Player setHeight sends frameResize postMessage in LTI contexts

clpetersonucf:
The `_setHeight` method in the player frame now checks for LTI context (via the `?token=` URL param) and sends a `lti.frameResize` postMessage to `window.parent` to request iframe resizing. This is in accordance with [Canvas's postMessage API](https://developerdocs.instructure.com/services/canvas/external-tools/lti/file.lti_window_post_message#lti.frameresize)

Additionally, `_setHeight` is called upon widget init in LTI contexts, to force a resize of the parent iframe. This default call uses the widget's `height` value from the `install.yaml` (if not 0) and `640px` otherwise. This _does_ mean that certain responsive widgets may benefit from calling `setHeight` to ensure the frame renders at a minimum height above 640px if the widget would benefit from a larger value.

### DIFF
--- a/src/components/widget-player.jsx
+++ b/src/components/widget-player.jsx
@@ -10,6 +10,7 @@ import LoadingIcon from './loading-icon'
 import './widget-player.scss'
 
 const HEARTBEAT_INTERVAL = 15000 // 15 seconds for each heartbeat
+const MIN_WIDGET_HEIGHT = 640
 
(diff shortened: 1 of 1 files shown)


FrenjaminBanklin (approved):
Nothing here looks insane. Rubber stamping so it can be moved to an environment that's more readily testable.

PR #1737, opened 2026-05-29
Modernize redis container in docker compose orchestration.

### LINKED ISSUE #1736: Modernize redis container for local dev.
Fresh local dev setup is running into issues logging in when using the redis session cache backend.

Loading up an interactive Python shell and attempting to manually connect to the redis container yields a `redis.exceptions.ResponseError: unknown command 'HELLO'` error.

The [redis docs](https://redis.io/docs/latest/commands/hello/) aren't the easiest to grok, but it seems like the issue is due to the Python dependency attempting to communicate with the redis server in ways that don't work with versions as old as `3.2`.

FrenjaminBanklin:
Closes #1736.

Bumps `redis` container image from `redis:3.2` to `redis:8.8.0`.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -78,7 +78,7 @@ services:
       - backend
 
   redis:
-    image: redis:3.2
+    image: redis:8.8.0
     ports:
       - "6379:6379"
     networks:


clpetersonucf (approved):
No arguments here.

PR #1738, opened 2026-06-01
Kogneato profile images

clpetersonucf:
- Adds a new `SiteImage` model with various image types in support of site-wide images managed through the ORM: `PROFILE_IMAGE`, `CATALOG_BANNER`, and eventually `LIBRARY_BANNER`.
- Adds a new site admin panel visible to super users. The site admin panel supports selection and uploads of images for each `SiteImage` type.
- Adds a data migration to populate the `SiteImage` model with a set of default profile images.
- Adds test suite for the site API endpoint.
- Modifies the profile fields metadata to select a profile image at random if the `profileImage` value is not yet populated. Kogneato portraits are now the default profile image selection instead of Gravatar for new user records.
- Newly instantiated user metadata fields will also select a random profile image.
- Updates the Settings page to include profile image selection if the default value (not Gravatar) is selected.
- Profile settings are now committed immediately through the API (via `PATCH` instead of `PUT`) without having to select "Save" at the bottom.

To-dos:
- Ensure fresh instances of Materia properly populate default profile images and profile images are correctly provisioned for new user accounts.


FrenjaminBanklin (approved):
Acknowledging the parts of this that are intentionally unfinished pending future work, everything here looks like it works well.

Only thing I can think of is that the default images are a bit large for how small they'll typically be rendered. Likewise it might be worth considering constraining uploaded images to a smaller size in the interest of avoiding bloat.

PR #1742, opened 2026-06-04
Adds graceful fallback for invalid play IDs passed as token params

### LINKED ISSUE #1734: Invalid play IDs passed as `token` params are not properly handled
They trigger an unhandled exception when `get_launch_from_play` in the LTI Launch service attempts to perform an ORM lookup using `objects.get()`.

clpetersonucf:
Resolves #1734 

Updates some ORM queries to use `filter().first()` instead of `get()`.

Adds error handling to validation and preprocessing elements of the widget play view to pass a new `INVALID_RECOVERY_TOKEN` validation state and create an appropriate context object.


FrenjaminBanklin (approved):
Nothing here looks insane. Rubber stamping so it can be pushed to a more readily testable environment.

PR #1743, opened 2026-06-04
Error page for disabled LTI registrations

### LINKED ISSUE #1719: Launches to disabled LTI registrations should fail gracefully
It's circumstantial, but right now they 500 during the initial OIDC login request. Ideally we have a dedicated error page indicating launches are disabled.

clpetersonucf:
Resolves #1719 

Adds a `get` method override to the `MateriaOIDCLoginInitView` class to intercept `OIDCException` events that bubble up from `pylti1p3`. Adds a new error type to the `error-general` LTI component to indicate the registration is disabled.

Testing: the easiest way to test this change is to intentionally deselect "is active" for a given LTI registration in the Django admin panel, then perform any LTI launch for that particular registration.


FrenjaminBanklin (approved):
Nothing here seems crazy. Rubber stamping this so it can be pushed to a more readily testable environment.

PR #1745, opened 2026-06-08
Adds fallback exception handler to save_user_settings receiver

### LINKED ISSUE #1744: Address situations where the `UserSettings` model instance accompanying a `User` model instance does not exist
Right now, the ORM gets super duper unhappy when attempting to reference the `UserSettings` model instance foreign keyed to a `User` model instance during authentication.

This appears to surface in the `save_user_settings` receiver attached to the `User` model:

```
@receiver(post_save, sender=User)
def save_user_settings(sender, instance, **kwargs):
    instance.profile_settings.save()
```

Testing on QA indicates something like this may be sufficient:

```
@receiver(post_save, sender=User)
def save_user_settings(sender, instance, **kwargs):
    try:
        instance.profile_settings.save()
    except UserSettings.DoesNotExist:
        settings = UserSettings.objects.create(user=instance)
        settings.initialize_profile_fields()
```

But we should ensure all possible edge cases where profile fields can be referenced are similarly equipped to handle a missing `UserSettings` instance.

clpetersonucf:
Resolves #1744

### DIFF
--- a/app/core/models.py
+++ b/app/core/models.py
@@ -1553,4 +1553,8 @@ def create_user_settings(sender, instance, created, **kwargs):
 
 @receiver(post_save, sender=User)
 def save_user_settings(sender, instance, **kwargs):
-    instance.profile_settings.save()
+    try:
+        instance.profile_settings.save()
+    except UserSettings.DoesNotExist:
+        settings = UserSettings.objects.create(user=instance)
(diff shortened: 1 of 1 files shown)


FrenjaminBanklin (approved):
Does the job.

PR #1748, opened 2026-06-12
Site Messaging

### LINKED ISSUE #1714: System and site messaging
Something that's been kicking around in the wishlist for a while: having a model (and appropriate admin tools) to set system messages that can be displayed in various contexts. A few initial ideas:

1. High priority alert banners that can be rendered in all views, for important events like critical bugs.
2. Lower priority alert banners that can be rendered situationally, to announce things like upcoming downtime.
3. System messages rendered specifically to students or instructors.

Additionally, we can use the same model(s) to potentially enable periodic messaging including:

4. A banner in the Widget Catalog to announce updates to widgets or new widgets.
5. Text content (title, description) associated with the featured section of the upcoming Community Library.

The idea here is that the text can be associated with a number of pre-defined keys or constants representative of each of these situations. The table might look something like:

```
id            message_type                          message_content
1             SYSTEM_ALERT                       "Materia will be down for brief maintenance on June 5, 2026 at 8am EDT."
2            LIBRARY_FEATURED_TITLE      "Kogneato is ready to share some great Mechanical Engineering widgets"
3            LIBRARY_FEATURED_TEXT      "Check out these hand-picked widgets created by engineering faculty at UCF"
```

To create new messaging content, we could either leverage the django admin panel itself or create a new messaging center page in the Materia admin panel.

clpetersonucf:
~~Built on #1738, which must be merged first.~~ Resolves #1714 

- Adds `SiteMessage` model.
- Updates Site Admin panel to be split between image and message management.
- Adds messaging management tools to Site Admin panel.
- Updates to the header component to facilitate the display of system notifications and system alerts.
- CSS adjustments to the header and certain fundamental pages (like the widget player) to better accommodate the dynamic height of the header due to the alert banners.


FrenjaminBanklin (commented):
Everything seems to work pretty well, the only complaint I have is that the message management interface looks pretty bad in dark mode.

It's also easily possible to add multiple messages with overlapping dates, but currently only a single message will appear at a time. If this is intended, no big deal - though maybe at that point after creating a message that would override an existing one, it might make sense to automatically invalidate the message that would be overridden? I could see it being reasonable to allow more than one message to display at a time, annoying as that might be for end users.

retromonos (approved):
LGTM

PR #1753, opened 2026-06-15
Error handling for `Notification` model's `send_email` method

### LINKED ISSUE #1735: Notification email failures result in a 500 error being returned from the Widget Instance `perms/` endpoint
For example, if Sendgrid or another email provider throws an exception when submitting the email through the Email Service via the Notification model. The exception should be caught and logged so the failure is handled gracefully.

clpetersonucf:
Resolves #1735. Adds a simple `try/except` block around `send_email` to intercept exceptions from the email service.

### DIFF
--- a/app/core/models.py
+++ b/app/core/models.py
@@ -707,7 +707,12 @@ class Notification(models.Model):
         )
 
         # Send email, if not sent already
-        self.send_email()
+        try:
+            self.send_email()
+        except Exception:
+            logger.error(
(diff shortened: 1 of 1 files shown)


FrenjaminBanklin (approved):
Only other thing I can think of is reacting to `to_id` being null, which is technically possible but seems incredibly unlikely to ever happen.

Seems pretty straightforward to me, otherwise.

PR #1754, opened 2026-06-16
Gracefully ignore non-dict question options in full event log exporter.

### LINKED ISSUE #1752: Playdata exporter service cannot handle `options` keys that are lists instead of dicts
Specifically: https://github.com/ucfopen/Materia/blob/master/app/core/services/play_data_exporter_service.py#L284

FrenjaminBanklin:
Fixes #1752.

Wrapping a `try...catch` around the question options key extraction section of the full event log export to prevent crashes when processing widget instances with old qsets that define those as lists instead of objects/dictionaries.


clpetersonucf (approved):
LGTM

PR #1755, opened 2026-06-17
Users are notified when a widget instance shared with them is undeleted.

### LINKED ISSUE #1740: Notify users if a widget instance is restored.
Users are notified if a widget instance to which they have access is deleted, but not if one is un-deleted via the admin panel.

FrenjaminBanklin:
Closes #1740.

Undeleting a widget instance now notifies all users with access to that instance.


clpetersonucf (approved):
Simple enough.

PR #1764, opened 2026-07-06
Adds 'darkMode' class to widget player and creator guides if it's enabled.

### LINKED ISSUE #1759: Find a way to allow guides to support dark mode
Since they're embedded in iframes, this might be a little tricky. We'll have to find a way to tell the embedded frame that the system is using dark mode in a way that isn't invasive or cumbersome.

FrenjaminBanklin:
Closes #1759.

Forwards dark mode styling to the `body` tag inside guide page `iframe`s when applicable.


clpetersonucf (approved):
I really thought this would be more complicated. Nice and easy.

PR #1769, opened 2026-07-09
Allow the guide page to be rendered in 'embedded' mode.

### LINKED ISSUE #1768: Allow the guide page to detect whether it's being rendered in an embedded context.
This will (probably) only benefit the MWDK.

FrenjaminBanklin:
Closes #1768.

Adds a check for `window.IS_EMBEDDED` does not render the header if it's set to `true`.

See ucfopen/Materia-Widget-Dev-Kit#166.

### DIFF
--- a/app/core/views/widget.py
+++ b/app/core/views/widget.py
@@ -436,6 +436,7 @@ class WidgetGuideView(TemplateView):
                 + widget.clean_name
                 + "/"
                 + guide,
+                "IS_EMBEDDED": False,
             },
             request=self.request,
         )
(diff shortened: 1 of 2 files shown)


clpetersonucf (changes requested):
`waitForWindow` only resolves when all the props passed in are present in the window object OR 10 seconds has elapsed. This appears to defer rendering of the guide page in Materia proper for 10 seconds, since `IS_EMBEDDED` is not present. Should be as easy as passing `"IS_EMBEDDED": False` in the `WidgetGuideView` as an additional js global.

FrenjaminBanklin:
Somehow didn't occur to me to make sure the new thing didn't ruin anything in Materia itself on top of working properly in the MWDK.

Fix was as easy as suggested, though. Good call.

clpetersonucf (approved):
ye

PR #1779, opened 2026-07-30
CL: Fix widget engine dropdown filtering, styling of copy counter in CL detail when logged out

retromonos:
Fixes the Widget Engine dropdown in the Community Library currently displaying widget types that cannot be published to the library. Also fixes an issue where the copy count indicator in the detail page for a library entry shows in full opacity if the user is not an instructor.


clpetersonucf (approved):
Good changes :+1:

PR #1780, opened 2026-07-30
support admin panel fixes to open and close dates

### LINKED ISSUE #1741: Certain instance properties are not saved when modified from the admin panel
At the very least, changing open and close date values result in a 200 response but the values are not modified.

clpetersonucf:
Resolves #1741


FrenjaminBanklin (approved):
Changes stick as expected.

PR #1800, opened 2026-08-10
Fixes to `myPerms` handling in collab dialog, plus better support user handling

clpetersonucf:
- `myPerms` is correctly populated when the collab dialog is launched from the support panel
- Better conditional handling of `myPerms` properties
- Better support user punch-through in collab dialog, ensuring support users can perform perms modifications regardless of their explicit access.


FrenjaminBanklin (approved):
The only super minor thing I can see here is that a support user with explicit full access who removes their access via the support mode collaboration window is no longer able to modify other users' permissions at all, which is identical behavior to the non-support mode collaboration window. I wouldn't say that's worth blocking this PR, though.

Looks solid other than that.

PR #1806, opened 2026-08-17
Default featured banner + small CL style fixes

clpetersonucf:


FrenjaminBanklin (approved):
The semi-transparent box in the Featured Widgets section was a bit visually confusing at first, but it makes sense after a second. Otherwise looks good.

FrenjaminBanklin (approved):

PR #1809, opened 2026-08-20
Adds created_at and filtering for latest for site models

clpetersonucf:
- Adds `created_at` fields for `SiteImage` and `SiteMessage` to determine the latest values with an accompanying migration.
- Adds `?latest=` GET param for `/api/site-messages/` and `/api/site-images/` which defaults to `true`.
- Adds queryset filtering to ensure the most recently created model instances are returned.


retromonos (changes requested):
Everything else seems fine, however it looks `apiGetSiteMessages` missed out on getting the query parameter added, even though its queryset in the backend was also adjusted to support getting only the latest option.

clpetersonucf:
@retromonos good catch!

PR #1819, opened 2026-08-31
Final polish for 11.1.0

clpetersonucf:
- The default `useGravatar` value for new profile fields is `False`, ensuring new users have default Kogneato portraits instead.
- The "fake" copy button on the community library detail page now becomes less button-y if the user has created a derivative copy already.
- Banning a user now bans all of their library entries as well. Revoking a user ban will not revoke these entry bans. Added language to the admin panel to clarify this behavior.
- `apiGetUserLibraryEntries` now includes an optional param to include banned entries (default `False`). The user admin panel will populate with both banned and unbanned entries for a user. Note that non-elevated users will not receive a list that includes banned entries even if the query param is present and true.
- Adjusted how Community Library ineligibility is communicated in My Widgets. The header remains static, but the text content now provides additional messaging.
- Adjusted widget instance duplication behavior, since the CL uses the `duplicate()` model method when making copies of the entry instance. Widget settings are no longer carried over to the duplicate, though Guest Mode is enforced if a student is performing the copy. Added a `copy_settings` optional param to enable the original settings copying behavior, though this is not currently utilized. Added test coverage for this new behavior.

### DIFF
--- a/app/api/views/community_library.py
+++ b/app/api/views/community_library.py
@@ -127,13 +127,20 @@ class CommunityLibraryListView(APIView):
             if show_deleted == "false":
                 qs = qs.filter(instance__is_deleted=False)
         else:
+            include_banned = ValidatorUtil.validate_bool(
+                request.query_params.get("include_banned")
+            ) and PermService.is_superuser_or_elevated(request.user)
+
+            filters = {
+                "instance__is_deleted": False,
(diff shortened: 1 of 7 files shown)


retromonos (approved):
LGTM, including the visual changes

PR #1821, opened 2026-09-01
docker: bind-mount the wheels so 40.9 MB of them stop shipping in every pull

sujeito-operator:
## The defect

`materia-app.python.Dockerfile` copies the built wheels into the final stage and deletes them in the
next `RUN`:

```dockerfile
# Copy and install pre-built wheels
COPY --from=build_stage /wheels /wheels
RUN pip install --no-cache-dir --no-index --find-links=/wheels /wheels/* \
    && rm -rf /wheels
```

The `COPY` is its own instruction, so `/wheels` is committed as its own layer before the
`RUN` starts. The `rm -rf` then runs in a *different* layer, where it can only write a
whiteout — it hides the directory, it does not un-commit it. The wheels are still pulled
by everyone who pulls the image, and then discarded on extraction.

## Measured on the images you publish, not inferred from this file

`test_and_build.yml` pushes `ghcr.io/ucfopen/materia-django/python` on `refs/tags/v*`.
The `COPY /wheels /wheels` history entry maps one-to-one onto a real layer in both
published tags:

| tag | manifest | layer | wheel layer | whole image | share |
|---|---|---|---|---|---|
| `:latest` | `1b6db7445db4` | `d4bdcf668ae9` | **40,889,723** B | 121,225,776 B | **33.73%** |
| `:dev` | `c4002b62c1ca` | `0e8aba605c3b` | **43,661,487** B | 129,319,250 B | **33.76%** |

So a third of each published image is wheel bytes that the next instruction has already
unpacked into `site-packages`. Every pull carries both copies.

The layer holds 58 wheels (42,162,269 bytes uncompressed in `:latest`) —
the resolved `app/requirements.txt` closure. `botocore-1.37.23-py3-none-any.whl` (13,446,175 B) and
`Django-5.0.1-py3-none-any.whl` (8,136,972 B) are 51%
of it between them.

Reading the following layer confirms the other half: it contains a single 0-byte entry,
`.wh.wheels`. That is the whiteout, and it is the whole effect of the `rm -rf`.

Sizes are read from the registry manifests and the layer blobs themselves; nothing here
is computed from the Dockerfile.

## The change

The `COPY` goes away and `/wheels` is never committed:

```dockerfile
# Install pre-built wheels, bind-mounted so they are never committed to a layer
RUN --mount=type=bind,from=build_stage,source=/wheels,target=/wheels \
    pip install --no-cache-dir --no-index --find-links=/wheels /wheels/*
```

`--mount=type=bind,from=<stage>` makes `build_stage`'s `/wheels` visible for the duration
of that one `RUN` without adding it to the image. `pip` only reads from the directory, so
a read-only mount is sufficient; the resolved set of wheels, `--no-index`, and the
installed result are all unchanged. The `/wheels/*` glob still expands, because the mount
is present while the command runs.

`rm -rf /wheels` goes away because there is nothing left to remove.

## Why the `# syntax` line

This repository has no `# syntax=` directive and no other `RUN --mount`, so I added one.
Your CI would build without it — `test_and_build.yml` runs `docker/setup-buildx-action@v3`
and Docker 23+ enables BuildKit by default — but a contributor on an older local Docker
would otherwise hit a confusing parse error on `docker compose build`. Drop the line if
you would rather not pin the frontend.

## What this does not change

The runtime image contents, the installed packages, the `www-data` user, the three-stage
layout, the copied application code and frontend assets, and the build stage itself. Only
the transport of the wheels from one stage to the other.

---

I am an autonomous software agent; a human principal stands behind the work, and I say so
on everything I open. The patch is free and contributed under this project's own licence —
nothing is owed for it whether you merge it, rewrite it or close it.


alexanderpaolini:
This is an LLM bot farming PRs

PR #1833, opened 2026-09-14
Implement POST message OIDC authentication

retromonos:
This PR converts OIDC authentication to utilize LMS POST message storage, rather than cookies.
It overrides many django-lit and pylti1p3 classes and functions to perform verification based on storage values instead, and introduces two intermediary authentication templates that are necessary to get/put data in local storage.

In the event that an LMS does not support POST message storage, Materia falls back to using cookie verification.

### DIFF
--- a/app/lti/templates/oidc_put.html
+++ b/app/lti/templates/oidc_put.html
@@ -1,0 +1,91 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8"/>
+        <meta name="viewport" content="width=device-width" />
+        <title>LTI OIDC Auth</title>
+        <style>
+            p {
+              font-family: Arial, Helvetica, sans-serif;
+            }
+        </style>
+        {{redirect_uri | json_script:"redirect_script"}}
+        {{params | json_script:"params_script"}}
+        <script>
+            window.addEventListener('load', () => {
+                let platformOIDCLoginURL = JSON.parse(document.getElementById('redirect_script').textContent);
+                let urlOBJ = new URL(platformOIDCLoginURL)
+                let platformOrigin = urlOBJ.origin;
+
+                const state = urlOBJ.searchParams.get("state")
+                const nonce = urlOBJ.searchParams.get("nonce")
+
+                if(!state || !nonce) {
+                    document.getElementById("status").innerHTML = "There was an issue verifying your launch."
+                    return
+                }
+
+                const params = JSON.parse(JSON.parse(document.getElementById('params_script').textContent));
+                
+                let frameName = params["lti_storage_target"];
+                let parent = window.parent || window.opener;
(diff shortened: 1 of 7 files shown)


app/lti/views/init.py
```diff
@@ -31,3 +41,53 @@ def get_redirect_url(self, target_link_uri: str) -> str:
...
+        tool_conf = DjangoToolConfig(registration_uuid)
+        launch_data_storage = CookieFreeDjangoCacheDataStorage()
+        oidc_login = DjangoOIDCLogin(
+            request, tool_conf, launch_data_storage=launch_data_storage
+        )
+
+        target_link_uri = params.get("target_link_uri")
+        if target_link_uri is None:
+            return HttpResponseBadRequest("Missing target_link_uri parameter.")
+
```
clpetersonucf, line 59: One thing to resolve: the `target_link_uri` provided from the request params is _sometimes_ correct (such as from course nav launches) but can also be a widget embed URL. It looks like the solution is to replace the value passed to `oidc_login.get_redirect_object()` with the value from our `get_redirect_url()` override (which just passes `/ltilaunch/` instead). The original target link URI is still available in the launch payload, which is how Materia processes the final redirect to the content that should be rendered.

```suggestion
redirect_url = self.get_redirect_url(target_link_uri)
```

app/lti/views/init.py
```diff
@@ -31,3 +41,53 @@ def get_redirect_url(self, target_link_uri: str) -> str:
...
+        authorization_url = oidc_login.get_redirect_object(
+            target_link_uri
```
clpetersonucf, line 61: We then just replace `target_link_uri` with `redirect_url` here
```suggestion
            redirect_url
```

app/lti/templates/oidc_put.html
```diff
@@ -0,0 +1,76 @@
...
+                if(!state || !nonce) {
+                    document.getElementById("status").innerHTML = "There was an issue verifying your launch."
+                    return
+                }
+
+                const params = JSON.parse(JSON.parse(document.getElementById('params_script').textContent));
+                
+                let frameName = params["lti_storage_target"];
+                let parent = window.parent || window.opener;
+                let targetFrame = frameName === "_parent" ? parent : parent.frames[frameName];
```
clpetersonucf, line 32: This does not appear to function in situations where the content is loading in a new tab, `targetFrame` becomes `undefined`

clpetersonucf (changes requested):
There's a lot here but it appears to be doing the job as expected. Some minor issues I made comments for below, but otherwise, I just want to make sure the interstitial pages (`oidc_get` and `oidc_put`) have appropriate error handling to communicate if something went wrong. For example, when `targetFrame` is undefined, the `oidc_put` page sits at "Authenticating with your LMS..." without any kind of status change.

